PR #5, opened 2024-11-11
chore(deps): disable bat dependency application features

tranzystorekk:
I noticed the `bat` dependency was grabbed as-is, which includes a lot of features and transitive dependencies unnecessary when using it in library mode.

This hopefully cleans this up (there's a noticeable drop in `Cargo.lock` entries)


alexpasmantier (approved):
Thanks for spotting that and for the PR 🙏🏻

PR #14, opened 2024-11-11
nix flake shell + rust-toolchain.toml setup

tukanoidd:
Hi, I really like this project and would like to use it. 
My main system is NixOS, so I have to either use nixpkgs or flakes, due to how things work there, won't delve too deep into it. This PR is for essentially packaging the application but also enables an environment shell with all the required setup to be able to work on the project itself (as I'm thinking of potentially contributing to this project in the the future (QOL improvements only most likely) if I find time). 
I am happy to maintain this setup if needed, although it should be stable for quite a while :)


alexpasmantier:
Hi, thanks for the interest and for this first contribution :)

Tbh I'm really not that versed into Nix configurations, and would be glad to have your support in maintaining this if you're willing to do it.

Btw I just made a couple of updates to the project's workspace configuration so can you test this PR still works on your system before we proceed?

Thanks!

tukanoidd:
> Btw I just made a couple of updates to the project's workspace configuration so can you test this PR still works on your system before we proceed?

Sure, I'll check

tukanoidd:
Hmm, it has issues with crates/television-*
image
as it thinks it's a directory name instead of glob-pattern, which is prolly a bug in the module im using for packaging

tukanoidd:
I'll use a different one, no need to do anything from your end

tukanoidd:
i think i set it up now, but need to wait before nix gets all the dependencies and stuff before i can check to be sure

tukanoidd:
Ok, seems good now, doing final test by adding it to my system config to make sure it works properly

tukanoidd:
yep, works, can be merged

tukanoidd:
btw, whenever you need to update Cargo.lock, please ping me so i can update the flake.lock, as it depends on that for getting right dependencies from registry, it's a quick cli command execution (`nix flake update`) so shouldn't take long to update

alexpasmantier (approved):

alexpasmantier:
Thanks for the follow-up! Merging this ✅

alexpasmantier:
@tukanoidd We should update the package.

Could this be done by a ci workflow automatically on each new release?

tukanoidd:
> @tukanoidd We should update the package.
> 
> Could this be done by a ci workflow automatically on each new release?

i dont know how that would be fixed by CI, since you need to update the files in the repository itself, and i dont know any bots that can create PRs lke that. From where should i update though? Just from main or are you making more changes in some other branch where i should run the update through?

tukanoidd:
Ok, sorry, my bad, i think I got confused a bit when initially wrote the message about needing to update with every Cargo.lock as i think i was thinking of a different framework at a time that required that (ive had a very hectic month). The flake needs to be updated mostly only when you change channel in rust-toolchain.toml. Atm all is good, i just updated it on my laptop without an issue

tukanoidd:
image 

:) (diregard the build date, it's always like that with nix)

tukanoidd:
I update my system fairly often (once in 2-4 weeks), so i would pick up on issues anyway, or others would create an issue if their flake stopped working, so I think to save you energy, just don't think about it too much and ping me in the issues about the flake if they arise, don't want you to feel burdened by this  :) Otherwise, i will most likely just create a PR with an update on my own whenever issues appear

PR #60, opened 2024-11-23
primitive version of Docker channel

bertrand-chardon:


crates/television-channels/src/channels/docker.rs
```diff
@@ -0,0 +1,125 @@
...
+        .await;
+    let summaries = match result {
+        Ok(image_summaries) => image_summaries,
+        Err(e) => {
+            panic!("{:?}", e)
+        }
+    };
+    summaries
+        .iter()
+        .filter(|f| f.repo_tags.len() > 0 && f.repo_digests.len() > 0)
```
bertrand-chardon, line 110: I access [0] below so that's for security but there's probably a better way to do this

crates/television-channels/src/channels/docker.rs
```diff
@@ -0,0 +1,125 @@
...
+        .for_each(|f| {
+            injector.push(
+                DockerImage {
+                    repository: f.repo_digests[0].clone(),
+                    tag: f.repo_tags[0].clone(),
+                    image_id: (f.id).clone(),
+                },
+                |_, cols| {
+                    cols[0] = (f.repo_digests[0]).clone().into();
+                },
```
bertrand-chardon, line 120: I am NOT clear about this at all 😬 don't really know what that does

PR #77, opened 2024-11-26
fix(output): quote output string when it contains spaces and points to an existing path

alexpasmantier:


crates/television/config/ui.rs
```diff
@@ -11,6 +11,7 @@ pub struct UiConfig {
     pub use_nerd_font_icons: bool,
     pub ui_scale: u16,
     pub show_help_bar: bool,
+    #[serde(default)]
```
fredmorcos, line 14: Out of curiosity, was this checked-in by mistake?

alexpasmantier: No, this actually fixes a feat I added yesterday. Just threw it in for conveniency since it's such a small change to avoid making a separate PR.

PR #83, opened 2024-11-27
remove unnecessary clone() calls

bertrand-chardon:
this PR removes a bunch of redundant calls to `.clone()` 

the list was obtained by running 

```
cargo clippy -- -W clippy::redundant_clone
``` 

on the source and removing the associated calls

- I ran `cargo test` to ensure tests are still OK
- I ran `cargo build --release` to rebuild the app and used it without any issue

### DIFF
--- a/crates/television/event.rs
+++ b/crates/television/event.rs
@@ -152,7 +152,6 @@ async fn poll_event(timeout: Duration) -> bool {
 impl EventLoop {
     pub fn new(tick_rate: f64, init: bool) -> Self {
         let (tx, rx) = mpsc::unbounded_channel();
-        let tx_c = tx.clone();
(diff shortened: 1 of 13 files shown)


crates/television/event.rs
```diff
@@ -152,7 +152,7 @@ async fn poll_event(timeout: Duration) -> bool {
 impl EventLoop {
     pub fn new(tick_rate: f64, init: bool) -> Self {
         let (tx, rx) = mpsc::unbounded_channel();
-        let tx_c = tx.clone();
+        let tx_c = tx;
```
alexpasmantier: I guess we can just remove that line entirely and use `tx` later on

bertrand-chardon: I was wondering about that actually, will do

alexpasmantier:
Marvelous!

PR #88, opened 2024-11-30
perf: adding a small local cache for icon colors

bertrand-chardon:
the rationale behind this PR is that there should on average _way_ more entries than the amount of different icon colors for a typical run of `tv` on a filesystem

as such, we should benefit from a small local cache where the result of `Color::from_str` is cached across the different entries

### DIFF
--- a/crates/television/ui/results.rs
+++ b/crates/television/ui/results.rs
@@ -8,6 +8,7 @@ use ratatui::widgets::{
     Block, BorderType, Borders, List, ListDirection, Padding,
 };
 use ratatui::Frame;
+use std::collections::HashMap;
 use std::str::FromStr;
 use television_channels::channels::OnAir;
 use television_channels::entry::Entry;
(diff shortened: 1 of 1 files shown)


crates/television/ui/results.rs
```diff
@@ -63,6 +64,19 @@ impl ResultsListColors {
     }
 }
 
+pub fn get_cached_color(
+    icon_color: &str,
+    color_cache: &mut HashMap<String, Color>,
```
bertrand-chardon, line 69: here, I wonder if `String` is necessary, it's possible that having a `&str` would be enough

crates/television/ui/results.rs
```diff
@@ -74,14 +88,18 @@ where
     'b: 'a,
 {
     let results_list_colors = results_list_colors.unwrap_or_default();
+    let mut color_cache: HashMap<String, Color> = HashMap::new();
```
bertrand-chardon, line 91: here, using a `HashMap` is the natural go-to but there might be a better way to achieve this in Rust using either a different type or some fancy crate, tell me if you know of any

PR #90, opened 2024-11-30
perf: remove unnecessary clone()

bertrand-chardon:
this is another change proposed by `cargo clippy -- -W clippy::redundant_clone`


alexpasmantier:
fixed in #75

PR #107, opened 2024-12-08
Add bench for build results list

bertrand-chardon:
This PR adds a benchmark for performance-critical function `build_results_list` from `television_screen::results`

This should allows `television` to track performances improvements or regressions across time for that function (and other in the future)


Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 4
+version = 3
```
bertrand-chardon, line 3: @alexpasmantier I am not sure I understand this

Cargo.lock
```diff
@@ -2364,7 +2528,7 @@ dependencies = [
  "crossterm",
  "indoc",
  "instability",
- "itertools",
+ "itertools 0.13.0",
```
bertrand-chardon, line 2531: this is now a specific version for some reason @alexpasmantier

Cargo.lock
```diff
@@ -3258,7 +3434,7 @@ version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b3644627a5af5fa321c95b9b235a72fd24cd29c648c2c379431e6628655627bf"
 dependencies = [
- "itertools",
+ "itertools 0.13.0",
```
bertrand-chardon, line 3437: same here

Cargo.toml
```diff
@@ -82,6 +83,13 @@ human-panic = "2.0.2"
 copypasta = "0.10.1"
 ansi-to-tui = "7.0.0"
 
+[dev-dependencies]
+criterion = "0.5.1"
+devicons = "0.6.11"
```
bertrand-chardon, line 88: `devicons` is required for the `FileIcons` inside of the benchmark

benches/results_list_benchmark.rs
```diff
@@ -0,0 +1,472 @@
...
+            icon: Some(FileIcon {
+                icon: '\u{e606}',
+                color: "#ffbc03",
+            }),
+            line_number: None,
+            preview_type: PreviewType::Files,
+        },
+    ];
+
+    c.bench_function("results_list_empty", |b| {
```
alexpasmantier: why `results_list_empty` here ?

Cargo.lock
alexpasmantier: I think it's because I'm compiling the project using the nightly channel. I wouldn't worry too much about it though, let's just leave it as it is now.

Cargo.lock
alexpasmantier: 0.13.0 is the latest version https://docs.rs/itertools/latest/itertools/

seems fine

benches/results_list_benchmark.rs
bertrand-chardon: it's an oversight, my bad. The name comes from the initial test of the benchmark that I ran against an empty list of entries. I'll change the benchmark function id accordingly.

alexpasmantier (approved):

PR #108, opened 2024-12-08
perf: merge contiguous name match ranges

bertrand-chardon:
in the entries that are used in the `build_results_list` function in `television-screen::results`, it seems like there the ranges returned are sometimes contiguous

since they're used to create `ratatui` spans, we want to reduce them to their minimal equivalent form

### DIFF
--- a/crates/television-channels/src/entry.rs
+++ b/crates/television-channels/src/entry.rs
@@ -65,11 +65,11 @@ impl Entry {
         self.value = Some(value);
         self
     }
-
+    #[allow(clippy::needless_return)]
     pub fn minimal_name_match_ranges(self) -> Option<Vec<(u32, u32)>> {
         // This method takes the existing `name_match_ranges`
         // and merges contiguous ranges into the minimal equivalent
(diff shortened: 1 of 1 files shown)


crates/television-channels/src/entry.rs
```diff
@@ -66,6 +66,31 @@ impl Entry {
         self
     }
 
+    pub fn minimal_name_match_ranges(self) -> Option<Vec<(u32, u32)>> {
```
bertrand-chardon, line 69: @alexpasmantier I don't know if having this on the `Entry` struct itself is the best course of action, please advise

crates/television-channels/src/entry.rs
```diff
@@ -65,6 +65,31 @@ impl Entry {
         self.value = Some(value);
         self
     }
+    #[allow(clippy::needless_return)]
```
bertrand-chardon, line 68: this is apparently needed for a reason that still eludes me

using an explicit seems frowned upon, but the code doesn't compile when I remove it, so I opted for this

alexpasmantier (approved):

PR #132, opened 2024-12-15
Add argument to start tv in another working directory

### LINKED ISSUE #111: open in other than the working directory
**Problem**
It appears that tv will only open the current working directory. Am I correct?

**Feature**
Would be nice to do 'tv /etc' for example, and open tv in the /etc directory.

**Examples**
`'tv /etc'`

**Additional context**

defigli:
A new CLI arg was added to start tv in another directory:
 -d, --working-directory <PATH>

Resolves #111

### DIFF
--- a/crates/television/cli.rs
+++ b/crates/television/cli.rs
@@ -40,6 +40,10 @@ pub struct Cli {
     #[arg(long, value_name = "STRING")]
     pub passthrough_keybindings: Option<String>,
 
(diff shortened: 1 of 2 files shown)


alexpasmantier:
Thanks a lot for the PR!

I made a couple of changes to make the target directory a positional argument instead which I think would feel more natural in most situations:

```sh
tv /etc

tv files /etc

...
```

PR #138, opened 2024-12-17
docs(readme): make channel names consistent everywhere

peter-fh:
I made this PR following some confusion I had as a first time user reading the documentation. I tried GitRepos, git-repos and finally had to call list-channels to find the actual name for gitrepos. I think making all of the references to channels consistent with the spelling/capitalization of the CLI helps reduce confusion.


alexpasmantier (approved):

alexpasmantier:
Thanks for the fix 🙏🏻

PR #188, opened 2024-12-30
fix: bump version to match with the release

chenrui333:
👋 seeing some version mismatch with the release build. 

relates to https://github.com/Homebrew/homebrew-core/pull/202798


alexpasmantier:
My bad, thanks for spotting that 🙏🏻

alexpasmantier (approved):

PR #195, opened 2024-12-31
fix(channels): list-channels in kebab-case

fannheyward:
```
tv gitrepos
Unknown channel: gitrepos
```

And `tv list-channels` result, before the fix:

```
tv list-channels
Builtin channels:
        env
        files
        gitrepos
        dirs
        text
        alias
```

After:

```
./target/debug/tv list-channels
Builtin channels:
        env
        files
        git-repos // HERE
        dirs
        text
        alias
```


alexpasmantier (approved):

alexpasmantier:
Thanks for spotting that and thanks for the fix! 🙏🏻

PR #204, opened 2025-01-01
fix: automatically create configuration directory if it doesnt exist

tulilirockz:
I've noticed that the latest release (0.85.0) failed to start up if `$HOME/.config/television` doesnt exist. This should force the directory creation if the user doesn't have it.

Before:
image

After:
image


tulilirockz:
It might also be good to create the data directory there too, what you think?

alexpasmantier (approved):

alexpasmantier:
Thanks for spotting that and for the PR 🙏

PR #228, opened 2025-01-06
add workspace lints

xosxos:
Hey, nice work. There's a lot of potential, and the idea of custom channels with custom run commands is very appealing.

I was looking at the source, but clippy pedantic lints threw a lot of warnings and I noticed the project did not have a linting profile. I threw together a profile here, but it does leave a fair amount of warnings in. Maybe some variation of this, or another linting profile could be considered to guide future contributors on style decisions. Cheers


alexpasmantier:
Hi, thanks for the kind words 🙏🏻 
This looks nice, just removed a couple of lints I feel better leaving on.

Thanks!

Issue #194, opened 2024-12-31
Sort or  Group "text" channel results.


cozycode:
**Problem**
When using the "text" channel to search within files across a directory tree, it can be challenging to make sense of the results because they appear in a random order.  Matches from each file are interleaved with results from other files and the line numbers within each file also appear in random order.

**Feature**
After entering a search phrase when using the "text" channel/mode, it would be nice if the result list would sort or group matches from within the same file. It might also be nice if the matches from the same file were ordered in the same order they appear in the file. Right now I am seeing matches from the same file interleaved with matches from other files.  

**Additional context**
It might also be nice to have a toggle to list each file with matches exactly once in the left side (with the match locations highlighted in the preview text on right) or to list each match as it does now.

xosxos:
This would be a great feature. A good implementation of this could also allow the git log entries to appear in chronological order.

zyoNoob:
Since you did mention this here, I will give whatever info I found while trying to fix this myself but gave up because of time constraints.

- FzF handles history in a unique way, I didn't study the exact approach but hey have separate flow for deciding the order of history, might be worth checking it out.
- Nucleo the library that TV uses for fuzzy matching currently doesn't supporting no-sort or sorted behavior as mentioned on their repository, but I did check some pull requests created by a contributor that does sort - TLDR direct sorting on time and score didn't work correctly, and top of that I didn't manage to integrate the call for sorted list in the TV code.
- I tried forking TV to do sorting on TV side of things but it doesnt work properly because TV fetches a batch of search results using nucleo and then renders them, so when ever i would scroll the results the search results would resort in the current pane. From what i figured is to reduce memory usage and compute overhead.

So long story short from my research, FzF uses a multi-condition sort for history and search on it, and for TV we might need feature additions to nucleo.

raylu:
> fzf `^r` command history search is sorted by last-used initially, but once you start typing to search it appears to be sorted by string length

I don't think that's right. it uses `--scheme history`: https://github.com/junegunn/fzf/blob/84e515bd6ecff0ed07d1373ff8e770f91d933b2e/shell/key-bindings.zsh#L135
and you can try
```sh
echo "prefix 1\nprefix 22222\ncow goes moo\nprefix 3" | fzf --scheme history
```
and see that typing `prefix` keeps 22222 in the middle

raylu:
[...]

PR #231, opened 2025-01-06
feat(cable) keep data input order

xosxos:
Hey, I had issues with the git log order getting shuffled in `tv git-log`.

If you want to add another dependency, `IndexSet` from [indexmap](https://crates.io/crates/indexmap), would do the same thing as the `HashSet` AND preserve order with roughly the same performance [rustc PR45282](https://github.com/rust-lang/rust/pull/45282). Also, `indexmap` is already included in the dependency tree, so this shouldn't really affect compile times either. Cheers

### DIFF
--- a/crates/television-channels/Cargo.toml
+++ b/crates/television-channels/Cargo.toml
@@ -29,6 +29,7 @@ strum = { version = "0.26.3", features = ["derive"] }
 lazy_static = "1.5.0"
 toml = "0.8.19"
 regex = "1.11.1"
+indexmap = "2.7.0"
 
 [lints]
 workspace = true
--- a/crates/television-channels/src/channels/cable.rs
+++ b/crates/television-channels/src/channels/cable.rs
@@ -1,7 +1,7 @@
 use color_eyre::Result;
+use indexmap::IndexSet;
 use lazy_static::lazy_static;
 use regex::Regex;
-use std::collections::HashSet;
 use tracing::debug;
 
 use crate::cable::{CableChannelPrototype, DEFAULT_DELIMITER};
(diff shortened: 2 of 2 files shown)


xosxos:
I think this also closes #194

alexpasmantier:
Very nice, that's helpful, thanks 🙏🏻 

A couple of tests and linting errors to fix (rebasing on main should do it for the linting issues) and let's merge this.

xosxos:
Sounds good, and oh yes sorry, I didn't run the tests 😅

xosxos:
Also, I assume that this use of `HashSet` is for deduplification, which is a bit problematic as It's potentially a noticeable performance hit to re-allocate there, and at the same time it's unnecessary for many applications such as logs etc. I guess the first thing would be to change the hash algorithm to a faster one and then down the line if it bottlenecks do something a bit more sophisticated.

alexpasmantier:
You're absolutely right.

https://crates.io/crates/rustc-hash should be more than enough here and we might see a noticeable performance improvement.

I do agree deduplicating here does not suit all purposes, but I felt it was nice to have in most cases. If it becomes more of a problematic issue, doing something more sophisticated down the line like you mentioned might be the way to go indeed.

alexpasmantier:
Fyi https://github.com/alexpasmantier/television/pull/237

Thanks for the suggestion 🙏🏻

alexpasmantier:
Ended up dropping the deduplication process altogether in https://github.com/alexpasmantier/television/pull/248 since it solves an issue and is always good for performance.

Closing this, thanks 🙏🏻

PR #266, opened 2025-01-10
Link to nucleo directly

archseer:
Makes the library a bit more discoverable, great to see it in use!


alexpasmantier:
Thanks for the PR 😊

Issue #296, opened 2025-01-20
How about releasing Windows assets in zip format instead of tar.gz?


kachick:
## **Problem**

Cannot register into winget-pkgs

## **Feature**

How about releasing Windows assets in zip format instead of tar.gz?

## **Additional context**

We can extract tar.gz in latest Windows Explorer, however some Windows tools do not support it.
For example winget-cli, when adding the release asset to winget-pkgs, winget-cli and manifest creators do not support tar.gz yet.

https://github.com/microsoft/winget-cli/issues/2899
https://github.com/russellbanks/Komac/blob/3179c6ff53a7b04a0ba9133fc5d97f26e00cbebc/src/file_analyser.rs#L56-L113

```console
> komac new --urls https://github.com/alexpasmantier/television/releases/download/0.9.4/tv-0.9.4-windows-x86_64.tar.gz
> Enter a GitHub token ********
> Package identifier: alexpasmantier.television
> Package version: 0.9.4
Error:
   0: Unsupported file extension: "gz"
```

PR #298, opened 2025-01-20
chore: add zip format in a Windows release assets

kachick:
Resolves GH-296 without replacing current tar.gz for backward compatibility

I have tested with similar code, then the workflow uploads files like https://github.com/kachick/television/releases/tag/0.999.999


alexpasmantier:
Neat! Thanks for the contribution 🙏🏻

kachick:
Thank you :zap: merging!

PR #303, opened 2025-01-23
fix: Add repaint command to the fish shell scripts

### LINKED ISSUE #300: Prompt corrupted after using ctrl+t
**Description**
After using tv to select something and returning to the prompt it is slightly corrupted.

**Example**
[asciicast](https://asciinema.org/a/riZfSYK6nEcTGE4Caej5eMKat)

**Expected behavior**
Prompt displays as normal

**Actual behavior**
Some of the last line is duplicated and it just looks wrong.

**Environment**
 - OS: Arch Linux 
 - Rust version: 1.84.0
 - Project version: 0.9.4
 - Shell: Fish
 - Happens using Kitty and alacritty, doesn't;t happen if I switch to bash

**Additional context**
Any other context about the bug here.

jscarrott:
The call ensures that the prompt restores properly and looks correct after using television. This fixes #300 for me locally.


alexpasmantier (approved):

alexpasmantier:
@jscarrott Neat! Thanks a lot for the patch 🙏🏻

PR #313, opened 2025-01-25
chore(terminal): custom shell keybindings

### LINKED ISSUE #268: Allow overriding or disabling Ctrl+T keybinding in shell
**Description**
When I execute in the shell `Ctrl+T`, I get this error:

```
Error:
   0: No channel found for prompt:
```

I guess this happens because the shell integration is expecting a command here. However, I use [fzf-file-widget](https://github.com/junegunn/fzf/blob/master/shell/key-bindings.zsh) as generic option for any command.

Would it be possible to override `Ctrl+T` keybinding? I don't find the way.

**Example**
Just press `Ctrl+T` in the shell.

**Expected behavior**
Television allowing changing shell keybinding, or at least disabling some particular keybindings.

**Actual behavior**
I get the described error.

**Environment**
 - OS: Ubuntu 24.04
 - Project version: 0.9.2

bertrand-chardon:
- allow custom keybindings through configuration
- this only works for `ctrl-xxx` keybindings for now (using anything else in the configuration will make `tv` fall back to `ctrl-t` and `ctrl-r`

Fixes #268

### DIFF
--- a/television/utils/shell/completion.bash
+++ b/television/utils/shell/completion.bash
@@ -23,5 +23,5 @@ function tv_shell_history() {
   fi
 }
 
-bind -x '"\C-t": tv_smart_autocomplete'
-bind -x '"\C-r": tv_shell_history'
+bind -x '"{tv_smart_autocomplete_keybinding}": tv_smart_autocomplete'
+bind -x '"{tv_shell_history_keybinding}": tv_shell_history'
--- a/.config/config.toml
+++ b/.config/config.toml
@@ -195,3 +195,11 @@ toggle_preview = "ctrl-o"
 
 # gitrepos channel
 "nvim" = "git-repos"
+
+[shell_integration.keybindings]
+# controls which key binding should trigger tv
+# for shell autocomplete
+"smart_autocomplete" = "ctrl-t"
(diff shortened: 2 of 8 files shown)


television/utils/shell/completion.bash
```diff
@@ -23,5 +23,5 @@ function tv_shell_history() {
   fi
 }
 
-bind -x '"\C-t": tv_smart_autocomplete'
-bind -x '"\C-r": tv_shell_history'
+bind -x '"{tv_smart_autocomplete_keybinding}": tv_smart_autocomplete'
+bind -x '"{tv_shell_history_keybinding}": tv_shell_history'
```
alexpasmantier, line 27: isn't line 2 (of the file) causing any formatting issues ?

```bash
local current_prompt="${READLINE_LINE:0:$READLINE_POINT}"
```

alexpasmantier: Ah nevermind just saw that "replace" above 👍🏻

alexpasmantier (approved):

alexpasmantier:
Very nice 🔥

PR #363, opened 2025-02-18
Image preview

I-Azy-I:
I initially didn't notice that an image previewer was already implemented but commented out—still, I wanted to finish mine! :)

It works almost instantly on my side. I tested it in different terminals and only noticed some slowness in the RustRover-integrated terminal.

So far, I’ve tested it with PNG, JPEG, ICO, GIF, and TIFF formats, and it works well. In theory, it should support all formats that the image crate can handle. I included them in the file list but commented out the ones I haven’t tested yet.

To optimize memory usage, images are resized to a maximum of 128x128 before being cached. I’m not really sure what the best size is, since the image gets resized again when rendered to fit the preview window. 

Let me know if you have any feedback! 🚀

ferris
poke
street


alexpasmantier:
Nice job! Thanks for the contribution!

Tested it quickly and it seems to mostly just work on my side which is really nice :) 

Will take a look at the code in more details this evening 🙏🏻

alexpasmantier:
Okay, so the code is quite neat to be honest. Good job (and thank you).

My observations/questions:

Was caching the rendered images a big performance improvement? 

I must say while I do admire the ingenuity I'm not too much of a fan of using the file previewer's cache to store `CachedImageData` itself containing a cache 😅 

My understanding is that you did that in order to avoid calling `DynamicImage.resize` too often (potentially resizing before inserting into the `FilePreviewer`'s cache, and then potentially resizing again before rendering).

While that does sound like a nice optimization, I feel it brings unnecessary complexity to the code which I'd like to avoid for future maintainers/contributors. 

I feel like we could avoid the `CachedImageData` struct entirely by:
- caching slightly larger images into the `FilePreviewer`'s cache (just enough to be downscalable to every *reasonable* terminal size) (for images larger than that)
- experiment without trying to cache rendered images and assess whether the performance is satisfactory

Feel free to correct the above as I might've misunderstood the code :)

television/utils/image.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+impl Widget for ImagePreviewWidget<'_> {
+    fn render(self, area: Rect, buf: &mut Buffer)
+    where
+        Self: Sized,
+    {
+        // I also put the render of the block to be similar has the preview paragraph, but I believe they are both useless since the block is already rendered
+        let inner = self.block.inner(area);
+        self.block.render(area, buf);
```
alexpasmantier:
```suggestion
        let inner = self.block.inner(area);
```

this does look unnecessary indeed

television/utils/image.rs
```diff
@@ -0,0 +1,255 @@
...
+                    *buf_cell = cell.clone();
+                }
+            }
+        }
+    }
+}
+#[derive(Clone)]
+pub struct CachedImageData {
+    image: DynamicImage,
+    inner_cache: Arc<Mutex<Option<DisplayedImage>>>,
```
alexpasmantier:
```suggestion
    /// A potentially downscaled version of the original `DynamicImage`
    image: DynamicImage,
    /// An inner cache destined to cache the final displayed image (ratatui-land)
    /// which might have been resized again to fit to the screen's dimensions
    inner_cache: Arc<Mutex<Option<DisplayedImage>>>,
```
just to make sure I understand this correctly

I-Azy-I:
Indeed, having two layers of caching, especially with the added complexity of using a mutex, feels unnecessary for a feature that doesn't provide significant benefits.

I believe you have a good understanding of the overall process :)

To summarize, the images are processed as follows:
1. The image is first opened and resized using a low-cost filter. Since precision is not critical at this stage, this step ensures that the image fits within the predefined maximum height and width constraints.
2. After conversion, it is stored in the preview cache.
3. When displaying the file, the image undergoes a second resizing step (this time using a higher-quality filter) before being converted into a format suitable for rendering.
Since step (3) is executed on every frame, I explored an optimization where the final resized image is cached to avoid performing this operation repeatedly.

Benchmarks & Findings
To measure the impact of caching, I added timing instrumentation in the draw_preview_content_block function within preview.rs:

```rust
let t = Instant::now();
let rp = build_preview_widget(
      // ...
);
f.render_widget(rp, inner);
info!("{:?}", t.elapsed().as_nanos());
```
(I hope it is a relevant to do it)

The collected benchmark results are summarized in the tables below:
|         | no cache + no widget 128x128 | no cache + no widget 512x512 | cache + no widget 128x128 | cache + no widget 512x512 | no cache + widget 128x128 | no cache + widget 512x512 | cache + widget 128x128 | cache + widget 512x512 |
|---------|------------------------------|------------------------------|---------------------------|---------------------------|---------------------------|---------------------------|------------------------|------------------------|
| Img 1 | 716817                       | 2.52753e+06                  | 377483                    | 341947                    | 490484                    | 2.4508e+06                | 24282                  | 26924                  |
| Img 2 | 429791                       | 2.11781e+06                  | 266526                    | 328321                    | 333766                    | 2.05183e+06               | 22917                  | 22700                  |
| Img 3 | 461366                       | 1.85112e+06                  | 230273                    | 268936                    | 305558                    | 1.80356e+06               | 19901                  | 20489                  |
| Img 4 | 480809                       | 2.05774e+06                  | 233010                    | 270287                    | 305618                    | 1.97572e+06               | 25021                  | 20048                  |
| Img 5 | 595371                       | 2.68458e+06                  | 341272                    | 310738                    | 446385                    | 2.4918e+06                | 27664                  | 20048                  |
> no widget means I convert it in Ratatui's Text to be rendered
> the size is correspond to the size after the first resize, so it is the size of the image in the preview cache

Initially, I used Lanczos3 as the filter type, which is ideal for cached or static images since they aren't resized frequently. However, to ensure a fair comparison, I also conducted tests with the Triangle filter. This alternative produces nearly identical visual results, as the images are already highly pixelated.

|       | Triangle filter + no cache + widget 128x128  |
|-------|----------------------------------------------|
| Img 1 | 253176                                       |
| Img 2 | 158800                                       |
| Img 3 | 214175                                       |
| Img 4 | 163124                                       |
| Img 5 | 383476                                       |

To provide further context, I also gathered performance metrics for various files:

| File                | Time   |
|---------------------|--------|
| main.rs             | 102377 |
| CONTRIBUTING.md     | 112632 |
| rust-toolchain.toml | 28340  |

Finally, I tested a static preview size:
|       | Static 50x100 |
|-------|---------------|
| Img 1 |20298   |
| Img 2 |15981   |
| Img 3 | 13239   |
| Img 4 | 26050   |
| Img 5 | 21986   |
> here, the size refers to the dimensions after the initial (and only) resizing step, which means it directly represents the size of the displayed image
In this case, the listed image size corresponds to the dimensions after the initial resize (and only resize), meaning it reflects the actual displayed image size.

I believe the dynamic sized image with no cache approach is the better option since, while it's slightly slower than a standard file, the difference is imperceptible to the end user. That said, the static version also performs well :) 

I'll commit all the version so you can try them

alexpasmantier:
@I-Azy-I Sorry for not coming back to you earlier, I've been a bit busy the last few days.

I was thinking that most of the complexity you had to come up with was in the end the symptom of somewhat of a design flaw, that of not being able to have any idea of what the screen might render from the "backend" side of the application (typically the previewers in your case).

With that in mind, I added a `UiState` attribute to the `Television` struct which contains information relative to the last rendered frame on screen.

This would allow, in your case, to know in advance what's your target area and thus decide how to resize the images before storing them into cache.

A next step (maybe in another PR) could be to "defend" against e.g. resizing the terminal by invalidating cache entries that aren't the right size etc.

What do you think?

I-Azy-I:
Wow, yeah, that would be perfect!
In fact, that's how I originally wanted to implement it, but after merging with the origin, it didn't work anymore (and looked a bit ugly).
Thanks for the new feature! I'll use it and push the new version :)

television/preview/previewers/files.rs
```diff
@@ -236,8 +250,63 @@ pub fn try_preview(
...
+                }
+            },
+            Err(e) => {
+                warn!("Error opening image {}: {:?}", entry.name, e);
+                None
+            }
+        };
+        if let Some(image) = option_image {
+            let preview_window_dimension = preview_window.map(|rect| {
+                (u32::from(rect.width - 2), u32::from(rect.height - 2)) // - 2 for the marge
```
alexpasmantier:
```suggestion
                (u32::from(rect.width.saturating_sub(2)), u32::from(rect.height.saturating_sub(2))) // - 2 for the margin
```

alexpasmantier:
This looks nice, should we merge it? 🥳

MadeInShineA:
Yes please!

PR #379, opened 2025-03-13
fix(zsh): use history command to include recent entries

keturn:
Problem: The `_tv_shell_history` command does not include the most recent entries in that shell's history.

It uses the `zsh-history` channel, but that's flawed because only the current shell knows its most recent history; it has not yet been written to `$HISTFILE`.

Instead, take input from the shell's history command.

### DIFF
--- a/television/utils/shell/completion.zsh
+++ b/television/utils/shell/completion.zsh
@@ -32,8 +32,7 @@ _tv_shell_history() {
 
     local output
 
-    output=$(tv zsh-history --input "$current_prompt" $*)
-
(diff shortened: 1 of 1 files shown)


keturn:
The same thing is true for bash too, but its `history` command has different flags.

alexpasmantier:
Thanks for the improvement, much appreciated 🙏🏻 

I want to change the preview pane default behavior with the stdin channel though before merging this so that it's disabled by default if no `preview-command` is specified.

alexpasmantier:
Done. Merging this 🎉

alexpasmantier (approved):

PR #383, opened 2025-03-14
Corrected git-repos name in channels.md

bpairet:
The channel git-repos was listed as gitrepos which made me think it was not working.


alexpasmantier:
Thanks for the fix 🙏🏻

PR #413, opened 2025-03-20
feat(shell): add fallback channel to the config for smart autocomplete

### LINKED ISSUE #321: Configurable default channel for fuzzy finding files
**Problem**
Apologies if this is already possible, I couldn't figure it out. 

Fzf has a cool feature where I can press ctrl+t in the middle of any command and it opens up a search UI where I can select a file and fzf spits out the file path where I had pressed ctrl+t. This is really useful across a variety of scenarios and to get this currently I need to have fzf (even though tv is more than capable of this). 

**Feature**
The feature would be fairly simple imo. tv should allow configuring a "fallback" channel (which in this case I would set to files) which would run when the command doesn't match any configured usages. This would allow using the files matcher (or whatever is configured) in arbitrary commands.

alexpasmantier:
Fixes #321

### DIFF
--- a/.config/config.toml
+++ b/.config/config.toml
@@ -109,6 +109,10 @@ toggle_preview = "ctrl-o"
 # E.g. typing `git checkout <CTRL-T>` will open television with a list of
 # branches to choose from.
 
+[shell_integration]
+# This specifies the default fallback channel if no other channel is matched.
+fallback_channel = "files"
+
 [shell_integration.channel_triggers]
 # Add your channel triggers here. Each key is a channel that will be triggered
 # by the corresponding commands.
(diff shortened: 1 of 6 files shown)


dryBoneMarrow:
This merge breaks the following config:
```
"" = "dirs"
```
Now when pressing `ctrl+t` with an empty prompt the fallback channel is used even though I configured it to use the dirs channels, this worked in older versions.  
I like having the files channel as default because most often it is the desired one. But for quickly changing into a directory it is really useful to use tv on an empty prompt. (This works without cd because I have `shopt -s autocd` set which changes the dir automatically)

PR #434, opened 2025-03-23
fix(filter): include hidden files

osamu2001:
Include hidden files in the search so that repositories starting with a dot, like `.config`, can also be found.

### DIFF
--- a/television/utils/files.rs
+++ b/television/utils/files.rs
@@ -75,6 +75,9 @@ pub fn walk_builder(
 ) -> WalkBuilder {
     let mut builder = WalkBuilder::new(path);
 
+    // Include hidden files
+    builder.hidden(false);
+
     // ft-based filtering
     let mut types_builder = TypesBuilder::new();
     types_builder.add_defaults();


alexpasmantier:
Hi, thanks for the PR.

I believe the default file channel behavior should be to exclude hidden files to avoid too much noise when starting tv in directories like $HOME, a git repo, etc.

The current way of adding the behavior you're after would be to:
- override the default files channel with something like
```toml
[cable_channel]
name = "files"
source_command = "fd . -I -t f"
preview_command = ":files:"
```
- create a dedicated channel (e.g. "hidden_files") to handle that use case if you still want the default "files" channel behavior

alexpasmantier:
@osamu2001 Added a small section [to the wiki](https://github.com/alexpasmantier/television/wiki/Tips-and-tricks#overriding-builtin-channels)

alexpasmantier:
You could do the following:

```toml
[[cable_channel]]
name = "git-repos"
source_command = "fd '\\.git$' ~/ --type d --no-ignore --hidden --format '{//}'"
preview_command = "cd {} && git log -n 200 --pretty=medium --all --graph --color"
```

alexpasmantier:
Closing the PR

PR #449, opened 2025-04-07
fix: silence the `string match` in tv_smart_autocomplete

dkmar:
The `string match` statement outputs strings if they match, so the `tv_smart_autocomplete` function was doing this if I invoke it with `'cd '` on my command line
image

It outputs `cd ` (in the first line) since the `$current_prompt` successfully matches (cause it ends with a space).

Let's fix this by making the `string match` silent with `-q`.


alexpasmantier (approved):

alexpasmantier:
Thanks for the fix 🙏

PR #450, opened 2025-04-07
feat(shell): support completion in the middle of a prompt

dkmar:
I think it's better to insert our output at the cursor so this can be invoked in the middle of a prompt. Previously we would assume the cursor is at the end of the prompt and invocation in the middle of one would discard everything after the cursor.


alexpasmantier:
Hi, thanks for the PR.

If that's the default behavior we're targetting, I'd like all shells to be consistent with it.

If you feel up to the task, have fun :) 

Otherwise I'll take this when I find the time.

Cheers

Issue #362, opened 2025-02-16
New option to specify header/name of the candidates list


asok:
**Problem**
I'm writing a script that will allow me to browse the filesystem and to choose file/directory. By using option `--passthrough-keybindings` I'm able to recursively call my script and go up or down in the filesystem tree. 
It would be nice to have the current directory printed instead of "STDIN".

**Feature**
Add a new option to specify the header/name of the candidates list.

**Examples**

This is my script:
```sh
#! /bin/sh

path="${1}/"

while :
do
output=$(ls -1pha ${path}. | sed 1,2d | tv --preview "if [[ -d ${path}{0} ]] ; then ls ${path}{0} ; else bat -n --color=never --line-range=:20 ${path}{0}; fi" --passthrough-keybindings "ctrl-h,ctrl-l,ctrl-g,esc")
    results=(${output// / })

    case ${results[0]} in
        "ctrl-l")
        if [[ -d "${path}${results[1]}" ]] ;
        then path="${path}${results[1]}" ;
        else
            output=${results[1]}
            break;
        fi
        ;;
        "ctrl-h")
        path="${path}../"
        ;;
        "ctrl-g")
        exit
        ;;
        "esc")
        exit
        ;;
        *)
        output=${results[0]}
        break
        ;;
    esac
done

echo "${path}${output}"
```

**Additional context**
This is how it looks in Zed editor:

![Image](https://github.com/user-attachments/assets/694b71ea-d2c8-4196-8d9c-9cf25236c631)

If I'm not mistaked `fzf` heas option `--header` which does something similar.

alexpasmantier:
Fixed by #472

PR #472, opened 2025-04-20
feat: add custom header for input field

nkxxll:
#362 

Add a custom header and the cli flag for the input field.

Maybe I should replace channel name with header name in the draw method instead of checking in the draw loop whether the header is there or not...


alexpasmantier:
The CLI option as well (that's the most important part actually 😅)

alexpasmantier:
image

Nice ! 🥳

PR #477, opened 2025-04-21
feat: add substring matching with exact flag

### LINKED ISSUE #368: A `--strict` or `--no-fuzzy-match` or `--exact-match` flag
**Problem**
there are some situations where you are searching through way too many files (for me, one application brings nearly 0.4 million files) and if search string becomes too long, I can feel the lag. Also I use it as my launcher for bookmarks, and they have urls which contain random url strings (for example youtube video urls) and many times they match with my fuzzy search string when i would not have them that higher in ranking.

**Feature**
a strict flag (something similar to `fzf`), wher if you use this flag, each search term is matched exactly. This is a (much) cheaper operation than general fuzzy matching (although both have the same lower asymptotic limit).

**Examples**
for `fzf` a example usage is `fzf --exact`

**Additional context**
Any other context about the feature request here.

nkxxll:
fixes #368 

@alexpasmantier I tried to parse down the exact flag to the find method, but I can't find an elegant way to do so because the channel creates its own configurations for the matcher inside of the `new` method. I started adding the exact flag to the new function of each of the channels but it just does not seem right to do so. Do you have an idea on how to improve the change? I would think of passing the config struct to the new function and exposing the config of the channel for channel transitions, but I'm not sure... 🤔 

I have done it here for the stdin channel just to show you how it would look like and it works fine😄! Looking forward to hearing your thoughts on that 👍 

ref issue: #368


alexpasmantier:
Maybe the most straightforward implementation in that case would be to short-circuit the `find` logic by adding that `if exact ...` condition earlier, e.g. [here](https://github.com/alexpasmantier/television/blob/83bc66a61e36fed39cb41280de7fbff565cc2025/television/television.rs#L186).

It's not the most elegant solution but it does the job and could be factored out into a `preprocess_pattern` function i.e.:

```rust
impl Television {
    ...
    fn find(&mut self, pattern: &str) {
        match self.mode {
            Mode::Channel => {
                self.channel.find(self.preprocess_pattern(pattern));
            }
            Mode::RemoteControl | Mode::SendToChannel => {
                self.remote_control.as_mut().unwrap().find(pattern);
            }
            Mode::Action => {
                todo!("Action mode not implemented yet");
            }
        }
    }
    ...
    fn preprocess_pattern(&self, pattern: &str) {
        ...
    }
}
```
what do you think?

television/television.rs
```diff
@@ -182,14 +197,35 @@ impl Television {
...
+                    Self::preprocess_pattern(self.matching_mode, pattern)
+                        .as_str(),
+                );
             }
             Mode::RemoteControl | Mode::SendToChannel => {
-                self.remote_control.as_mut().unwrap().find(pattern);
+                self.remote_control.as_mut().unwrap().find(
+                    Self::preprocess_pattern(self.matching_mode, pattern)
+                        .as_str(),
+                );
```
alexpasmantier: I don't think we want to apply the same logic to other modes (the remote control should stay fuzzy and not be impacted by the `--exact` flag).

alexpasmantier:
We can also add an integration test where we launch the app with a pre-filled input and the new `exact` option and check that it selects the right entry.

You can find examples in https://github.com/alexpasmantier/television/blob/main/tests/app.rs and use the app's event channel to send down instructions and control the application ([example](https://github.com/alexpasmantier/television/blob/main/tests/app.rs#L122-L126)).

nkxxll:
- removed the preprocessing for remote
- added one positive and one negative integration test for the exact flag
  - if you enter "fie" with fuzzy you should get both entries as matches with exact you get none
  - if you enter "file" you should get both entries as matches

alexpasmantier:
Neat 👍🏻 
Thanks for the PR, keep up the good work! 😊

PR #480, opened 2025-04-24
fix(results): remove keymap hint if help is disabled

### LINKED ISSUE #471: Help Shortcut Hint Displays When Run With --no-help Flag
**Description**

If you run `tv --no-help` the hint for how to open the help panel ("help: \<Ctrl-g>") displays at the bottom of the screen. Pressing Ctrl-g doesn't do anything when run with `--no-help`. The hint should probably be hidden.

**Example**

```sh
tv --no-help
```

**Expected behavior**

The help hint (Ctrl-g) at the bottom of the screen should not display.

**Actual behavior**

The help hint (Ctrl-g) shows at the bottom of the screen.

**Environment**

- OS: Linux (Pop!_OS)
- Project version: [e.g. 0.11.7]

nkxxll:
fixes #471

### DIFF
--- a/television/config/ui.rs
+++ b/television/config/ui.rs
@@ -6,11 +6,13 @@ use super::themes::DEFAULT_THEME;
 
 const DEFAULT_UI_SCALE: u16 = 100;
 
+#[allow(clippy::struct_excessive_bools)]
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Hash)]
 #[serde(default)]
 pub struct UiConfig {
     pub use_nerd_font_icons: bool,
     pub ui_scale: u16,
+    pub no_help: bool,
     pub show_help_bar: bool,
     pub show_preview_panel: bool,
     #[serde(default)]
@@ -25,6 +27,7 @@ impl Default for UiConfig {
         Self {
             use_nerd_font_icons: false,
             ui_scale: DEFAULT_UI_SCALE,
+            no_help: false,
             show_help_bar: false,
             show_preview_panel: true,
             input_bar_position: InputPosition::Top,
--- a/television/screen/results.rs
+++ b/television/screen/results.rs
@@ -253,8 +253,12 @@ pub fn draw_results_list(
     help_keybinding: &str,
     preview_keybinding: &str,
     preview_togglable: bool,
+    no_help: bool,
 ) -> Result<()> {
-    let mut toggle_hints = format!(" help: <{help_keybinding}> ",);
+    let mut toggle_hints = String::new();
+    if !no_help {
+        toggle_hints.push_str(&format!(" help: <{help_keybinding}> ",));
+    }
     if preview_togglable {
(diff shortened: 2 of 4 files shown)


television/config/ui.rs
```diff
@@ -6,11 +6,13 @@ use super::themes::DEFAULT_THEME;
 
 const DEFAULT_UI_SCALE: u16 = 100;
 
+#[allow(clippy::struct_excessive_bools)]
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Hash)]
 #[serde(default)]
 pub struct UiConfig {
     pub use_nerd_font_icons: bool,
     pub ui_scale: u16,
+    pub no_help: bool,
```
alexpasmantier, line 15: I believe this flag is already available just below (the `show_help_bar` flag)

television/config/ui.rs
```diff
@@ -25,6 +27,7 @@ impl Default for UiConfig {
         Self {
             use_nerd_font_icons: false,
             ui_scale: DEFAULT_UI_SCALE,
+            no_help: false,
```
alexpasmantier, line 30:
```suggestion
```

television/screen/results.rs
```diff
@@ -253,8 +253,12 @@ pub fn draw_results_list(
     help_keybinding: &str,
     preview_keybinding: &str,
     preview_togglable: bool,
+    no_help: bool,
 ) -> Result<()> {
-    let mut toggle_hints = format!(" help: <{help_keybinding}> ",);
+    let mut toggle_hints = String::new();
+    if !no_help {
+        toggle_hints.push_str(&format!(" help: <{help_keybinding}> ",));
+    }
```
alexpasmantier, line 261: I think this should be the only real required change (along with passing the actual flag to the context).

alexpasmantier:
Thanks for the PR! 
Made a couple of comments about the fact that the flag you're adding already exists but apart from that this looks good 👍🏻

television/config/ui.rs
nkxxll: Thought the same thing, it is not though, show_help_bar holds the state of whether the help bar is currently shown: if no_help is set the show_help_bar option is set to false and cannot be turned on because the keymap is not handled. However if no_help is not set then show_help_bar is also false until you hit c-g which flips it to true and turns "on" the view, else I could have used it...

nkxxll: for reference here the toggle https://github.com/alexpasmantier/television/blob/a602dda34758f9f4a24f1c77b589216c12b9cfba/television/television.rs#L649-L654

alexpasmantier: Ah yes you're absolutely correct, my bad 🙇🏻

alexpasmantier:
let's go 🚀 
thanks for the PR 🙏🏻

television/config/ui.rs
nkxxll: np ❤️

Issue #444, opened 2025-04-01
tv alias goes to the background


0ther0ne:
**Description**
when I run `tv alias`
the job goes to the background
rather than on the terminal

**Example**
```
$ tv alias
[1]+  Stopped              tv alias
$ jobs
[1]+  Stopped              tv alias
```

**Environment**
 - OS: Arch Linux
 - Rust version: 1.85.0
 - Project version: 0.11.5
 - bash shell

Rick-Phoenix:
Same issue here. If I launch tv first and then select the alias picker I don't have the same problem. It's only when it's launched with "tv alias" directly.

alexpasmantier:
Thanks for the feedback 🙏🏻 

Fixed in #484

Rick-Phoenix:
Thank you!

alexpasmantier:
@cr4ftx fyi

PR #484, opened 2025-04-28
fix(alias): move terminal raw mode before loading bat assets #444

cr4ftx:


television/tui.rs
```diff
@@ -40,7 +40,6 @@ where
     }
 
     pub fn enter(&mut self) -> Result<()> {
-        enable_raw_mode()?;
```
alexpasmantier: What happens if we leave this here? Is it some sort of no-op if raw mode is already activated?

The `enter` method is also called further down inside `resume` so it's probably important that we re-enable raw mode as some point after the app has been suspended.

cr4ftx: Hey @alexpasmantier, I've just tried to not remove the `enable_raw_mode` from Tui and it still works, but I wonder what is the use case of resume as crtl-z is disabled ?

alexpasmantier: Ctrl-z can still be handled by the application itself.

And it's important we keep the ability to suspend/resume from within the app for upcoming features such as performing actions on entries that spawn another interactive process (e.g. opening an entry with an external editor like vim).

swann-castel: I totally agree, but for now the application does not handle crtl-z keybindings right ?

We should keep this in mind for the future of television, I will investigate on why enabling raw mode trigger the process suspension after loading bat assets

cr4ftx: I totally agree, but for now the application does not handle crtl-z keybindings right ?

We should keep this in mind for the future of television, I will investigate on why enabling raw mode trigger the process suspension after loading bat assets

alexpasmantier: > I totally agree, but for now the application does not handle crtl-z keybindings right ?

No indeed

But spawning interactive processes (and so suspending and resuming) is just around the corner.

alexpasmantier:
Thanks !! 🥳 
Merging this

Issue #438, opened 2025-03-25
Respecting currently typed path for files channel


JordanCamirand:
**Problem**
Not respecting currently typed path for files channel
ex: `mv ~/Downloads/<CTRL T>`

**Feature**
When a path is typed to use that as the starting point for the files channel instead of the current directory

nkxxll:
i would love to do that pr but I kind of don't know where to start.
https://github.com/alexpasmantier/television/blob/dbff3a330b169c422ae384e373b934dceb8e01b2/television/main.rs#L174-L183
Somehow we would have to give the path to the files channel after checking whether the last argument given to the `--autocomplete-prompt` flag is a path (could use Path::new("...").exists() 🤔 ). Currently the path is determined before calling `Channel::new(paths)`. I guess we would have to pass it to the `new` method somehow.
@alexpasmantier do you have any ideas?

alexpasmantier:
I feel like we probably shouldn't prioritize this right away and instead work towards having television being able to directly plug into each shell's completion system.

Each shell has its own autocompletion mechanism and lots of completion scripts already exist (take a look at the zsh plugin ecosystem for instance).

Shells often provide a way of plugging a custom application to display autocompletion results and I believe that's the right way to do it for television.

Needless to say this would make the current shell integration feature redundant in a lot of scenarios but we could fall back to it for cases when no completion is available.

A good place to draw inspiration from might be [zsh's scripts](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh).

This probably isn't trivial and for having had a look at it I know zsh's autocompletion system is quite complex but it's feasible.

I know @cr4ftx was also interested by the subject.

Cheers

alexpasmantier:
Implemented for Fish in #494

alexpasmantier:
Closing this as completed.

PR #494, opened 2025-05-02
feat(shell): improve fish completion system

lalvarezt:
this pr adds the possibility of starting television from a different directory and it gives the option of complete a partial match. For example

```fish
cd ~/.config/<ctrl-t>    # opens television at '~/.config/' with an empty search bar
cd ~/.config/fi<ctrl-t>    # opens television at '~/.config/' with 'fi' in the search bar
```

solves (for fish shell) https://github.com/alexpasmantier/television/issues/438


alexpasmantier:
Great job, thanks for the PR 🙏🏻 
Merging this

alexpasmantier (approved):

PR #495, opened 2025-05-04
refactor!(previews): drop builtin previewers and all related code and dependencies

alexpasmantier:


cable/unix-channels.toml
lalvarezt, line 1: what about windows channels?

alexpasmantier: @lalvarezt I've updated windows channels as well: https://github.com/alexpasmantier/television/blob/main/cable/windows-channels.toml

I'm thinking we could use https://github.com/uutils/coreutils if windows lags behing.

What are your thoughts on this?

lalvarezt: It's not in the PR hence the question.

On the other topic since both `eza` and `bat` are available for windows using `winget`
```cmd
winget install --id=eza-community.eza -e
winget install --id=sharkdp.bat -e
```
I'd say for know it's not really an issue.

Note: When I'm in windows I use WSL but I've tested both commands natively and they work just fine

alexpasmantier: The windows channels were changed in https://github.com/alexpasmantier/television/pull/479 which wasn't released yet and will part of 0.12.

lalvarezt: there seems to be a misunderstanding, in this PR you're dropping the `:files:` preview from the `unix-channel.toml` file and the rust code that goes with it

image

the `windows-channels.toml` file still references this

image

there's no change in this PR adjusting the windows version of the file. Or did I misread the rust changes?

alexpasmantier: Ah yes you're correct!

This PR is still a work in progress and the windows previews do need updating 😊

PR #496, opened 2025-05-04
feat(orientation): add orientation for television #489

### LINKED ISSUE #489: Vertical Layout
**Problem**
Vertical layout (preview on top, search results on the bottom) is far more space-efficient on my monitor/usual window layout

**Feature**
Support different layouts

**Examples**
e.g. this is my Telescope layout with live-grep:

Image

cr4ftx:
Fixes #489 

This PR adds support for television orientation, input_bar_position is preserved and activating/deactivating the preview works for both :

1. Landscape (default orientation)

a. input_bar_position = "top"

image
image

b. input_bar_position = "bottom"

image
image

2. Portrait (new orientation)

a. input_bar_position = "top"

image
image

b. input_bar_position = "bottom"

image
image

### DIFF
--- a/television/draw.rs
+++ b/television/draw.rs
@@ -241,12 +241,13 @@ pub fn draw(ctx: &Ctx, f: &mut Frame<'_>, area: Rect) -> Result<Layout> {
         &ctx.tv_state.spinner,
         &ctx.colorscheme,
         &ctx.config.ui.custom_header,
+        &ctx.config.ui.input_bar_position,
     )?;
 
-    if layout.preview_window.is_some() {
+    if let Some(preview_rect) = layout.preview_window {
         draw_preview_content_block(
             f,
-            layout.preview_window.unwrap(),
+            preview_rect,
             &ctx.tv_state.preview_state,
             ctx.config.ui.use_nerd_font_icons,
             &ctx.colorscheme,
--- a/television/screen/input.rs
+++ b/television/screen/input.rs
@@ -6,11 +6,15 @@ use ratatui::{
     },
     style::{Style, Stylize},
     text::{Line, Span},
-    widgets::{Block, BorderType, Borders, ListState, Paragraph},
+    widgets::{
+        block::Position, Block, BorderType, Borders, ListState, Paragraph,
+    },
     Frame,
 };
 
 use crate::screen::{colors::Colorscheme, spinner::Spinner};
+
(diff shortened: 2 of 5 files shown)


cr4ftx:
Maybe we could change the tv logo orientation too, that would be a little easter egg

alexpasmantier:
This looks neat!

I'm thinking maybe we should also change that `files` header on the input bar to be at the bottom instead of on top when `input_bar_position` is set to `bottom`.

What do you think?

cr4ftx:
> This looks neat!
> 
> I'm thinking maybe we should also change that `files` header on the input bar to be at the bottom instead of on top when `input_bar_position` is set to `bottom`.
> 
> (both for portrait and landscape modes)
> 
> What do you think?

Do you want this only for the input title or all titles ?

alexpasmantier:
> Do you want this only for the input title or all titles ?

Only on the input bar should be fine

daniel-hernandez-sony:
It would be great if the orientation could be changed at runtime. But it might be easier to do that on a follow-up PR

television/draw.rs
```diff
@@ -241,12 +241,13 @@ pub fn draw(ctx: &Ctx, f: &mut Frame<'_>, area: Rect) -> Result<Layout> {
         &ctx.tv_state.spinner,
         &ctx.colorscheme,
         &ctx.config.ui.custom_header,
+        &ctx.config.ui.input_bar_position,
     )?;
 
-    if layout.preview_window.is_some() {
+    if let Some(preview_rect) = layout.preview_window {
```
alexpasmantier, line 247: 👍🏻

television/screen/input.rs
```diff
@@ -25,13 +29,18 @@ pub fn draw_input_box(
...
     let input_block = Block::default()
         .borders(Borders::ALL)
         .border_type(BorderType::Rounded)
         .border_style(Style::default().fg(colorscheme.general.border_fg))
-        .title_top(
+        .title_position(match input_bar_position {
+            InputPosition::Top => Position::Top,
+            InputPosition::Bottom => Position::Bottom,
+        })
+        .title(
```
alexpasmantier, line 43: 👍🏻

alexpasmantier (approved):

alexpasmantier:
@daniel-hernandez-sony great idea, I agree that should be done in another PR down the road.

PR #499, opened 2025-05-08
fix(layout): double check whether preview is enabled

nkxxll:


alexpasmantier:
Nicely spotted, thanks for the PR 🙏🏻

Issue #393, opened 2025-03-17
Impossible to paste when using shell integration with zsh


justinrlle:
**Description**
When running `tv` with the `ctrl+t` shortcut in a `zsh` shell, I can't paste some text in the prompt.

**Example**
With this simple `config.toml`:
```toml
[shell_integration.commands]
"cd" = "dirs"
```

When I type `cd` then `ctrl+t`, and then type `cmd+v`, nothing happens.

Happens also when using the `ctrl+r` history integration.

**Expected behavior**
I expect the content of my clipboard to be inserted in the prompt.

**Actual behavior**
Nothing get inserted

**Environment**
 - OS: MacOS
 - Rust version: same as releases, installed with Homebrew
 - Project version: 0.10.9
- shell: ZSH
- terminal: iTerm2 and Terminal.app

**Additional context**
A few things I tried:
- outside of shell integration, by just running `tv`, it works: I can copy paste.
- in bash, I had an error for the `ctrl+t` (surely linked to the version of bash that MacOS ships with), but I can paste in correctly for the `ctrl+r` history integration
- I tried to run `zsh` without any startup config (`zsh -f -d -i`), following this source: <https://unix.stackexchange.com/questions/579104/start-interactive-zsh-without-running-any-configuration-files-like-zshrc>, but it still didn't work, neither with `ctrl+t` nor `ctrl+r`.

So I think it's a bug in the zsh integration. Note that `fzf` does work correctly on that side, so I think it's doable, but I don't have a lot of deep knowledge of shells and integrating with them. Happy to help debug the issue or try out some fixes though!

---

And last but not least, thanks for your work on television, I love it!

alexpasmantier:
This is surprising indeed. I'm actually able to reproduce on arch linux and zsh. 
Will investigate, thanks for the feedback 🙏🏻

dkmar:
Looks like running `unset zle_bracketed_paste` fixes this

alexpasmantier:
@dkmar do you want to take this? Haven't been able to look into it yet.

dkmar:
Haven’t forgotten about this btw. I looked at other projects and have some notes. I’ll write an update comment on monday

alexpasmantier:
no problem! thanks for the note 😊

alexpasmantier:
This looks perfect and fixes the issue on my side.

Thanks for taking the time to make a contribution 😊

PR #512, opened 2025-05-15
fix(shell): Paste not working in zsh shell integration

kapobajza:
Fixes https://github.com/alexpasmantier/television/issues/393

### DIFF
--- a/television/utils/shell/completion.zsh
+++ b/television/utils/shell/completion.zsh
@@ -1,6 +1,22 @@
+_disable_bracketed_paste() {
+    # Check if bracketed paste is defined, for compatibility with older versions
+    if [[ -n $zle_bracketed_paste ]]; then
+        print -nr ${zle_bracketed_paste[2]} >/dev/tty
+    fi
+}
+
+_enable_bracketed_paste() {
+    # Check if bracketed paste is defined, for compatibility with older versions
+    if [[ -n $zle_bracketed_paste ]]; then
+        print -nr ${zle_bracketed_paste[1]} >/dev/tty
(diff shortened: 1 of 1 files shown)


television/utils/shell/completion.zsh
```diff
@@ -1,7 +1,23 @@
+_disable_bracketed_paste() {
+    # Check if bracketed paste is defined, for compatibility with older versions
+    if [[ -n $zle_bracketed_paste ]]; then
+        print -nr ${zle_bracketed_paste[2]} >/dev/tty
```
forivall, line 4: on my machine, i needed to change this to

```suggestion
        print -nr ${zle_bracketed_paste[2]} >${TTY:-/dev/tty}
```

television/utils/shell/completion.zsh
```diff
@@ -1,7 +1,23 @@
...
+    fi
+}
+
+_enable_bracketed_paste() {
+    # Check if bracketed paste is defined, for compatibility with older versions
+    if [[ -n $zle_bracketed_paste ]]; then
+        print -nr ${zle_bracketed_paste[1]} >/dev/tty
```
forivall, line 11: ditto

```suggestion
        print -nr ${zle_bracketed_paste[1]} >${TTY:-/dev/tty}
```

television/utils/shell/completion.zsh
alexpasmantier: Thanks for the feedback!
Would you mind opening a PR with the suggested changes?
Thanks 🙏

Issue #520, opened 2025-05-21
default_channel throws error in config.toml


different-name:
**Description**
Using the `default_channel` field in the main `config.toml` file throws the following error:
```
Error: Error parsing configuration file: /home/different/.config/television/config.toml

╔══════════════════════════════════════════════════════════════════════════════╗
║                                                                              ║
║  If this follows a recent update, it is likely due to a breaking change in   ║
║  the configuration format.                                                   ║
║                                                                              ║
║  Check https://github.com/alexpasmantier/television/releases/latest for the  ║
║  latest release notes.                                                       ║
║                                                                              ║
╚══════════════════════════════════════════════════════════════════════════════╝

Caused by:
    TOML parse error at line 1, column 1
      |
    1 | default_channel = "nixpkgs"
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    unknown field `default_channel`
```

**Example**
`~/.config/television/config.toml`
```toml
default_channel = "nixpkgs"
```

**Expected behavior**
To be able to set the default channel, as per
https://github.com/alexpasmantier/television/blob/7bbf5388984ea4a9afa4daad695add4c258c0fb1/.config/config.toml#L20

**Actual behavior**
Using the `default_channel` results in a config parsing error

**Environment**
 - OS: NixOS
 - Rust version: 1.86.0
 - Project version: 0.11.9

kapobajza:
It seems like `default_channel` is a config setting from a version that hasn't been released yet. If take a look at the [code from the latest version (0.11.9 and also the one you're using)](https://github.com/alexpasmantier/television/blob/0.11.9/.config/config.toml), you will see that it doesn't contain the `default_channel` setting.

lalvarezt:
in any case I don't think that option is being used in the code

https://github.com/alexpasmantier/television/blob/7bbf5388984ea4a9afa4daad695add4c258c0fb1/television/config/mod.rs#L44-L46

https://github.com/alexpasmantier/television/blob/7bbf5388984ea4a9afa4daad695add4c258c0fb1/television/channels/prototypes.rs#L78

The one I use in my config is 

https://github.com/alexpasmantier/television/blob/7bbf5388984ea4a9afa4daad695add4c258c0fb1/.config/config.toml#L106-L108

PR #521, opened 2025-05-22
fix(channels): use default_channel from config

lalvarezt:
sorry, first rust code ever 😅
solves https://github.com/alexpasmantier/television/issues/520.

With the current implementation the `default_channel` value from config was never loaded

### DIFF
--- a/television/channels/prototypes.rs
+++ b/television/channels/prototypes.rs
@@ -4,6 +4,7 @@ use std::{
     ops::Deref,
 };
 
+use crate::config::default_channel;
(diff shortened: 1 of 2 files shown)


alexpasmantier:
Thanks for the PR 🙏🏻 

While this solves the problem, I feel like it reveals a design flaw. We shouldn't have to load the configuration file twice and we can avoid that by delaying the moment we determine the channel to use.

Let me draft another PR and ping you when it's ready so that you may review it and give me your thoughts!

lalvarezt:
Closing in favor of #524

PR #525, opened 2025-05-25
feat(shell): improve zsh completion system

lalvarezt:
this pr adds the possibility of starting television from a different directory and it gives the option of complete a partial match. For example

```fish
cd ~/.config/<ctrl-t>    # opens television at '~/.config/' with an empty search bar
cd ~/.config/fi<ctrl-t>    # opens television at '~/.config/' with 'fi' in the search bar
```

solves (for zsh shell)
https://github.com/alexpasmantier/television/issues/438


alexpasmantier:
Very cool and nicely documented. Thanks for the PR 🙏🏻

PR #533, opened 2025-06-09
refactor(cable)!: cable format redesign

alexpasmantier:


lalvarezt:
@alexpasmantier you changed the name of the keybinding (now it's cycle_through_sources), the config file committed doesn't match

cable/unix/files.toml
```diff
@@ -26,6 +26,8 @@ quit = ["esc", "ctrl-c"]
 select_next_entry = ["down", "ctrl-n", "ctrl-j"]
 select_prev_entry = ["up", "ctrl-p", "ctrl-k"]
 confirm_selection = "enter"
+toggle_source = "ctrl-t"
+reload_source = "ctrl-r"
```
lalvarezt: this is now called cycle_through_sources

television/screen/keybindings.rs
```diff
@@ -236,18 +256,50 @@ fn build_keybindings_table_for_channel<'a>(
         colorscheme.mode.channel,
     ));
 
+    // Toggle source (optional)
+    let toggle_source_row = keybindings
+        .bindings
+        .get(&DisplayableAction::CycleThroughSources)
+        .map(|toggle_source_keys| {
+            Row::new(build_cells_for_group(
+                "Toggle source",
```
lalvarezt: same here, the new name is cycle_through_sources (maybe Cycle through sources?). so it matches with the config name

cable/unix/files.toml
alexpasmantier: Thanks for pointing that out!

Should now be fixed 👍🏻

television/screen/keybindings.rs
alexpasmantier: Same here

cable/unix/env.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "env"
+description = "A channel to select from environment variables"
+
+[source]
+command = "printenv"
+output = "{split:=:1..}" # output the value
+
+[preview]
+# output the value and make it easier to read for colon-separated values
+command = "echo '{split:=:1..|join:|split:\\::..|join:\n}'"
```
lalvarezt: another cheaper alternative, in either case they introduce anomalies for example if you have urls and such

```toml
# output the value and make it easier to read for colon and comma-separated values
command = "echo '{replace:s/[:|,]/\n/g|split:=:1..}'"
```

lalvarezt: also your template has a side effect

```sh
string-pipeline "{split:=:1..|join:|split:\\::..|join:\n}" "PATH=a=1:b=2:c=3"
a1
b2
c3
```

it should be

```sh
string-pipeline "{split:=:1..|join:=|split:\\::..|join:\n}" "VAR=a=1:b=2:c=3"
a=1
b=2
c=3
```

for path variables it doesn't matter, but for others if might be an issue

cable/unix/text.toml
```diff
@@ -1,39 +1,14 @@
...
 name = "text"
 description = "A channel to find and select text from files"
-# requirements = ["rg", "bat"]
+requirements = ["rg", "bat"]
 
 [source]
 command = "rg . --no-heading --line-number"
-display = "[{split:\\::..2}]\t{split:\\::2}"
-output = "{split:\\::..2}"
+display = '[{split:\::..2}]	{split:\::2}'
```
lalvarezt:
```toml
display = '[{split:\::..2}]	{split:\::2..}'
```

cable/unix/env.toml
alexpasmantier: Didn't mean to commit that, was just experimenting locally 😬

cable/unix/text.toml
alexpasmantier: same here, rolled back the change 👍🏻

cable/unix/env.toml
alexpasmantier: Thanks for pointing that out 🙏🏻

alexpasmantier:
@lalvarezt if you're looking for things to do, adding cli support for some of the prototype fields could be awesome 😁 
```sh
tv --source-command "..." --source-display "..." --source-output "..." \
    --preview-command "..." --preview-offset "..." --layout "portrait" \
    --preview-header = "..."
```

lalvarezt:
> @lalvarezt if you're looking for things to do, adding cli support for some of the prototype fields could be awesome 😁
> 
> ```shell
> tv --source-command "..." --source-display "..." --source-output "..." \
>     --preview-command "..." --preview-offset "..." --layout "portrait" \
>     --preview-header = "..."
> ```

sure, I can work on that

alexpasmantier:
Also added sort of a micro testing framework to spawn tv inside a pty and test various user workflows if you're curious to take a look 😊

alexpasmantier:
Adding the flags above to the cli would allow to test rather simply if the different formatting options work (which would have failed with your mod on the matcher yesterday)

lalvarezt:
> Adding the flags above to the cli would allow to test rather simply if the different formatting options work (which would have failed with your mod on the matcher yesterday)

yup, and they'll open the door on-the-fly channels (with some tweaks they'll make for a good batch testing system, end-to-end)

lalvarezt:
@alexpasmantier what's the next step?

I was planning on cleaning up some more the cli interface and writing the docs for that module

alexpasmantier:
That sounds perfect!

I'd like to do the following things before the release:

- rework the remote control UI (maybe show it as a popup instead of coming in from the side so we can split it into multiple panes and display the channels as well as a pane for the corresponding requirements if any with a check that they're available on the user's PATH and the associated prototype description)
- rewriting the channel documentation from the ground up
- writing a small transition guide from 0.11 to 0.12 for existing users explaining what's changing
- never enough tests as the maintenance perimeter is growing
- maybe squeezing in actions per channel but I feel more confident doing that in a later release so we can take a bit more time to make something nice and stable

If you think of anything else, feel free to mention it! :-)

alexpasmantier:
> - rework the remote control UI (maybe show it as a popup instead of coming in from the side so we can split it into multiple panes and display the channels as well as a pane for the corresponding requirements if any with a check that they're available on the user's PATH and the associated prototype description)

Done in [8c0fa82](https://github.com/alexpasmantier/television/pull/533/commits/8c0fa82c767c0107b6c4d3c9f9cb888c0506fa52).

Screenshot From 2025-06-22 16-31-58

alexpasmantier:
Note to self: more integration tests for cross-features compatibility (e.g. reloading a channel while a multi selection is active, cycling through sources while selection is active, etc.).

lalvarezt:
> Also added sort of a micro testing framework to spawn tv inside a pty and test various user workflows if you're curious to take a look 😊

I just had a look, this is awesome 👏

alexpasmantier:
We're left with the following before finally releasing 0.12:
- [x] rewriting the channel documentation from the ground up (in a simple and concise manner)
- [ ] writing a small transition guide from 0.11 to 0.12 for existing users explaining what's changing (this can be more or less similar to the 0.12 release note)
- [ ] never enough tests as the maintenance perimeter is growing (especially testing feature interactions like selecting entries and then reloading the channel etc.)

Are there any items you want to work on @lalvarezt or should I just take these?

lalvarezt:
> We're left with the following before finally releasing 0.12:
> 
>     * [ ]  rewriting the channel documentation from the ground up (in a simple and concise manner)
> 
>     * [ ]  writing a small transition guide from 0.11 to 0.12 for existing users explaining what's changing (this can be more or less similar to the 0.12 release note)
> 
>     * [ ]  never enough tests as the maintenance perimeter is growing (especially testing feature interactions like selecting entries and then reloading the channel etc.)
> 
> 
> Are there any items you want to work on @lalvarezt or should I just take these?

I'm doing the final touches on batch of test, focused on cli. I'm really liking the ptytester you've built (you might consider creating a crate for other ratatui apps, it's a really cool idea).

After that I'll think about other tests

lalvarezt:
getting an error with this last commit on WSL
```
cargo run -- --cable-dir ./cable --config-file ./.config/config.toml env
```

it uses `env.toml` from the `windows` directory.
image

What is the expected logic? Should the user use directly the `unix` directory for example when doing the custom --cable-dir? like so

```
cargo run -- --cable-dir ./cable/unix/ --config-file ./.config/config.toml env
```

this makes sense in a way, that you can have any structure you like inside cable folder and unix/windows is just the default convention. On the flip side we have hardcoded those paths.

alexpasmantier:
For WSL, it should be 
```
tv --cable-dir cable/unix
```
The `cable/windows` being specifically for a windows terminal running powershell.

That's only a way of separating the collections inside the repo so that users of each collection can browse the channels on GitHub and download them using:
```
tv --update-channels
```

But this doesn't presume anything regarding each user's own cable dir which can be organised freely.

alexpasmantier:
Merging this into main as the "cable refactoring" part is behind us and it doesn't really make any sense to keep the rest in this monolithic PR.

I'll be adding a few last commits to main and will release 0.12 soon after.

PR #536, opened 2025-06-11
POC of changes

lalvarezt:
## text cable update
there was a minor bug in the text cable template

before
`display = "[{split:\\::..2}]\t{split:\\::2..}"`

after
`display = "[{split:\\::..2}]\t{split:\\::2}"`

this allows to capture items after the split that also contain the separator

## artifact when rendering the preview with debug flag enabled

this is POC you can solve it in another way

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,21 +33,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
-name = "android_system_properties"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
-dependencies = [
- "libc",
-]
-
-[[package]]
 name = "anes"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -226,21 +211,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
-name = "chrono"
-version = "0.4.41"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c469d952047f47f91b68d1cba3f10d63c11d73e4636f24f08daf0278abf01c4d"
-dependencies = [
- "android-tzdata",
- "iana-time-zone",
- "js-sys",
- "num-traits",
- "serde",
- "wasm-bindgen",
- "windows-link",
-]
-
-[[package]]
 name = "ciborium"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -358,12 +328,6 @@ dependencies = [
  "once_cell",
  "windows-sys 0.59.0",
 ]
-
-[[package]]
-name = "core-foundation-sys"
-version = "0.8.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
 name = "cpufeatures"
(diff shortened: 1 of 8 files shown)


lalvarezt:
On a second thought, perhaps the best is to just break compatibility and add a debug parameter to format. I don't like the cloning part for nothing
What do you think?

alexpasmantier:
> before
> display = "[{split:\\::..2}]\t{split:\\::2..}"
> 
> after
> display = "[{split:\\::..2}]\t{split:\\::2}"
> 
> this allows to capture items after the split that also contain the separator

It's the other way around right? 😛 

Not sure I understand the debug part though.

I feel like the debug option shouldn't be a part of the template itself but should rather be a cli flag (for the cli application) or a boolean flag for the lib's entrypoints (parse, format).

lalvarezt:
perhaps and example would be better.
in the text channel you use

```toml
[source]
command = "rg . --no-heading --line-number"
display = "[{split:\\::..2}]\t{split:\\::2}"
output = "{!split:\\::..2}"
```

manually running
`rg . --no-heading --line-number`

you see
```
television/screen/preview.rs:183:        )
television/screen/preview.rs:184:        .padding(Padding::new(0, 1, 1, 0));
television/screen/preview.rs:186:    let inner = preview_outer_block.inner(rect);
television/screen/preview.rs:187:    f.render_widget(preview_outer_block, rect);
television/screen/preview.rs:188:    Ok(inner)
television/screen/preview.rs:189:}
```

notice :: after padding

image

on the other hand

```toml
[source]
command = "rg . --no-heading --line-number"
display = "[{split:\\::..2}]\t{split:\\::2..}"
output = "{!split:\\::..2}"
```

shows the correct value

image

lalvarezt:
> Not sure I understand the debug part though.
> 
> I feel like the debug option shouldn't be a part of the template itself but should rather be a cli flag (for the cli application) or a boolean flag for the lib's entrypoints (parse, format).

half and half 😅. Currently you can do both, there's the option to use the set_debug or with_debug to override the functionality (but due to the way you pass the info in your code I'm forced to clone to modify). 

On the other hand it is possible to set the ! sign at the beginning of the template to request the debug information to be shown, that way I could debug the output (`output = "{!split:\\::..2}"
`) by inspecting both the result and the debug information when selecting an item (it's quite handy)

the problem lies in that you serialize all (display and output) in the same way so we cannot block the display template from showing it. and it causes the rendering issue as it shows the debug info on the screen. usually you won't see it, but if a user does it by mistake the it will look weird.

a summary would be, we don't want to allow display to show debug info, even if it has ! (it can be done with set_debug(false) or chain .with_debug() and we want to be able to show debug if the user wants for the output like below

image


is it clearer now, maybe

edit: sure you could add a flag in the cli for tv to show debug for output and then when when parsing just go directly for set_debug(whatever the value is) and then you have control to hide it for display and show it if needed for output

edit2: display and preview are in the same boat, debug is not wanted there

lalvarezt:
updated the parser to allow to pass the debug option when parsing, this gives you control very early
and proposed another solution for to disable the debug during preview without cloning

lalvarezt:
this latest change disables debug for preview.command, preview.offset, and source.display and
leaves it enabled for source.output. So this config works as it should

```toml
[source]
command = "rg . --no-heading --line-number"
display = "[{!split:\\::..2}]\t{split:\\::2..}"
output = "{!split:\\::..2}"

[preview]
command = "bat -n --color=always {!split:\\::0}"
env = { "BAT_THEME" = "ansi" }
offset = "{!split:\\::1}"
```

alexpasmantier:
The first point makes perfect sense, I got it (the `split:\\::2..` instead of the `split:\\::2`). Was just referring to the pull request description which I think has it the other way around 👍🏻 

EDIT: here's the actual PR description

> there was a minor bug in the text cable template
> 
> before
> display = "[{split:\\::..2}]\t{split:\\::2..}"
> 
> after
> display = "[{split:\\::..2}]\t{split:\\::2}"
> 
> this allows to capture items after the split that also contain the separator

alexpasmantier:
> and we want to be able to show debug if the user wants for the output like below

That's where I tend to disagree. I feel like `debug` should entirely be disabled by default when you're using the lib as a dependency for a larger project. You don't want opt-out printing to stdout or stderr, you want these kind of features as opt-ins.

Also, I feel like `debug` shouldn't be sending to stderr/stdout but should instead be sending to a more classical configurable destination (typically a log file), and be enabled using rust native features like compilation profiles (dev, release, ...). A nice and easy way to achieve this could be to rely on something like https://docs.rs/tracing/latest/tracing/.
If you agree on the above, feel free to take a look at https://github.com/alexpasmantier/television/blob/main/television/logging.rs for inspiration.

lalvarezt:
I see, you have a point there. Then it makes sense to make the debug system inert by default and you can opt-in later on (if you choose) when the rest of the details have been worked out. I'll prepare a new release tomorrow. 
Feel free to close this PR it was just for documentation purposes

lalvarezt:
new version published FYI, in `0.12.0` all debug information is suppressed by default with `parse(...)`, if you use `parse_with_debug(..., None|Some(true)|Some(false))` you'll have the option choose if you want to see the debug messages or not (for now they're still with `stderr`)

PR #539, opened 2025-06-13
fix(app): honor cli no-help and no-preview

lalvarezt:
precedence should be 
cli flags > local config > global config


alexpasmantier:
Thanks for spotting that 🙏🏻

alexpasmantier:
Merging and adding tests in a separate commit.

PR #540, opened 2025-06-13
New channel and added reload and toggle actions

lalvarezt:
Add env channel
Add support to toggle between sources and reload them, this enables the following

```toml
[source]

[keybindings]
toggle_source = "ctrl-t"
reload_source = "ctrl-r"
```

now you can have the same channel with two different representations, for example with and without hidden files

the change is backwards compatible, if you put a simple string nothing changes

### DIFF
--- a/cable/unix/env.toml
+++ b/cable/unix/env.toml
@@ -1,0 +1,39 @@
+[metadata]
+name = "env"
+description = "A channel to select from the environment variables"
+
+[source]
+command = "printenv"
+interactive = false
+# env
+display = "{}"           # show the key-value pair
+output = "{split:=:1..}" # output the value
+
+[preview]
+# command = "cut -d= -f2 <<< ${0} | cut -d\" \" -f2- | sed 's/:/\\n/g'"
+command = "echo '{split:=:1..}'" # output the value
+
+[ui]
+layout = "landscape"
+ui_scale = 100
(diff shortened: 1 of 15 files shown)


alexpasmantier:
Love the idea of having a `reload` functionality.

Not too sure about multiple sources for a single channel but let's give it a try! It does solve the "enabling hidden files" on-demand for fd usecase and might be useful for the future.

Do you mind if I iterate on this PR and make a couple of changes? I'd like to change the implementation a bit and typically avoid having to re-create the prototype each time you want to cycle through your sources.

lalvarezt:
> Love the idea of having a `reload` functionality.
> 
> Not too sure about multiple sources for a single channel but let's give it a try! It does solve the "enabling hidden files" on-demand for fd usecase and might be useful for the future.
> 
> Do you mind if I iterate on this PR and make a couple of changes? I'd like to change the implementation a bit and typically avoid having to re-create the prototype each time you want to cycle through your sources.

please do, you have a better understanding of the ins and outs 😉

alexpasmantier:
Refactored it a bit, and everything seems to be working fine :-)

Cable channel downloading from github is broken (because of the way we serialize keybindings) but that shouldn't be too hard to fix and will be done back on the original PR.

Merging this 🎉

PR #545, opened 2025-06-16
Preview size customization

lalvarezt:
This PR gives support to define multiple preview sizes and to toggle them at will

### DIFF
--- a/benches/main/ui.rs
+++ b/benches/main/ui.rs
@@ -450,6 +450,7 @@ pub fn draw(c: &mut Criterion) {
                     false,
                     false,
                     false,
+                    Some(50),
                     false,
                     cable.clone(),
                 );
--- a/cable/unix/files.toml
+++ b/cable/unix/files.toml
@@ -11,7 +11,4 @@ command = "bat -n --color=always {}"
 env = { BAT_THEME = "ansi" }
 
 [ui]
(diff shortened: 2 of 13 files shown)


lalvarezt:
not sure why the CI test is failing, #ItWorksOnMyPC 😅. I guess you had the same issue with your previous commit

alexpasmantier:
Great work!

Being able to adjust the preview size directly for each channel is a feature I've been wanting to add for a certain time.

I do feel that having multiple possible sizes for a single channel is a bit of a premature optimization though and we'd be better off starting with something simple (i.e. specifying a single preview size per channel). You can still keep that improvement stashed away so we can come back to it if we ever realize that we need the extra feature.

This option should also ideally be reflected in the CLI and configuration file. Let me know if you need any help with that! 🙏🏻

alexpasmantier:
PS: fixed the failing CI test, rebasing the PR should do the trick 😊

lalvarezt:
made the changes as you suggested. 

I do really like the option of having multiple options for preview but I guess that hiding the preview essentially gives us two options to play with. Perhaps it is too much 🤔, we can revisit this later

alexpasmantier:
Awesome feature 😁 

Thanks for the PR! 🎉

PR #547, opened 2025-06-16
UI customization

lalvarezt:
This PR adds support for customizing

* input_header
* preview_header
* preview_footer

all the items are of type MultiTemplate, this gives some flexibility when showing the selected value, let's say on the preview header, if it's long or you want only a section


television/config/ui.rs
```diff
@@ -8,6 +9,34 @@ use super::themes::DEFAULT_THEME;
...
+{
+    let raw: Option<String> = Option::<String>::deserialize(deserializer)?;
+    match raw {
+        Some(s) => Template::parse(&s)
+            .map(Some)
+            .map_err(serde::de::Error::custom),
+        None => Ok(None),
+    }
+}
+
```
alexpasmantier: [Don't these already exist?](https://github.com/alexpasmantier/television/blob/133b3789b059b88afc070e1c044156a7a9a6440c/television/channels/prototypes.rs#L131-L143)

alexpasmantier: (they're just wrongly named, I'll update that 😬 )

lalvarezt: yes, there was the command version and didn't want to mess with those and change the name

alexpasmantier: Ah okay, feel free to merge the two then :)

alexpasmantier:
Also I'm thinking we should add a few tests for all this new behavior. Tell me if you need a hand with that 😄

lalvarezt:
> Also I'm thinking we should add a few tests for all this new behavior. Tell me if you need a hand with that 😄

yes, please. I'm not too familiar with the testing structure you have in place, I would appreciate it 🙇‍♂️

alexpasmantier:
No problem!

Should I merge the two functions too or are you doing that?

lalvarezt:
> No problem!
> 
> Should I merge the two functions too or are you doing that?

done, we should have a much more clear ser/deser structure

alexpasmantier:
> done, we should have a much more clear ser/deser structure

Ha, nice, well done!

alexpasmantier:
Rebased onto the original PR and added a couple of tests.

**Side note:** I feel like there's something wrong about how we handle CLI overrides and the whole thing should be refactored into a simpler form. But that's not linked to this PR in particular and can be done afterwards.

alexpasmantier:
Merging this, thanks for the PR 🙏🏻

Issue #352, opened 2025-02-08
Add a post-process command and allow user to specify default post process per channel and to enable/disable and preview the results.


rtaycher:
**Problem**
For stuff like docker or processes, etc its usual to first filter and then pipe it t9 something like `| awk '{print $1}'`
Sure you can get some of the information sometimes by using preview like the docker images command does but it doesn't let you quickly filter by something other then selection (the docker images channel has you go down a list of ids)
You may want selection to be docker image/process id or process id or file name from ripgrep but filter by some sort of name or detail.

Technically you can add this pipe outside TV command but that adds extra typing and having to type it before you filter with TV. TV and default postprocess commands  would make this really cool for a lot of commands (docker images/ps/docker PS/git/etc)

**Feature**
Add a command to be run after TV selection. 
Allow it to be previewable and modifiable. You can reuse the preview window or better yet allow a smaller window above preview window since it would only need a few lines of output to show it was working properly.
Allow user to modify it inside the UI.
It could either be a command or position and a separator (might be faster and would take care of most use cases although some might still prefer piping to another command)

**exampe**

[[cable_channel]]
name =
source_command =

Add

post-process = "awk '{print $1}'"

or 

post-process = {pos=1, sep=" "}

 Or both?

anlek:
Yeah, I'm finding that I need the output to be slightly different then the list view name... `post_process` or something similar would really help

makzsslm:
This would be really really cool cool ! I've been thinking whether I could use television as an app launcher like Rofi/Alfred/Raycast. 

Adding a post-processing step, you could do all kinds of things like:
- Make an app launcher channel that opens apps
- Make web search channel with autosuggestions that opens the browser
- Make a calculator channel that copies the result to the clipboard
- ...

While this might be doable by analyzing the output, doing it in the channel config would be way easier.

alexpasmantier:
I'm planning on implementing https://github.com/alexpasmantier/television/issues/16 in the week ahead.

I feel like the feature that is being asked for in this issue could be achieved directly with custom actions on completion without the need for a special extra feature.

**Example:**

For a given channel, a custom action might be to:
- open the selected entry in vim, 
- open that entry in your browser, 
- maybe `cd` into the selected directory 
- etc. 
 
And I'm assuming that you could, just as with the cases above, take tv's output and re-pipe it into tv's stdin while applying the transformation to each element with a (somewhat involved) shell command(/script) as well.

Something like:
```sh
# action: transform
[...]

PR #549, opened 2025-06-17
Transition channel POC

lalvarezt:
@alexpasmantier this is still a work in progress, would like to get your opinion. Is this how you envision the functionality?

### DIFF
--- a/.config/config.toml
+++ b/.config/config.toml
@@ -95,6 +95,8 @@ toggle_preview = "ctrl-o"
 reload_source = "ctrl-r"
 # Cycle through the available sources for the current channel
 cycle_sources = "ctrl-s"
+transition = "f2"
+transition_back = "f1"
 
 
 # Shell integration
--- a/cable/unix/dirs.toml
+++ b/cable/unix/dirs.toml
@@ -4,7 +4,8 @@ description = "A channel to select from directories"
 requirements = ["fd"]
 
 [source]
-command = ["fd -t d", "fd -t d --hidden"]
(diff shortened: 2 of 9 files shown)


lalvarezt:
I was also considering the config like:

```toml
[transition]
# remove trailing /
files = ". {replace:s/\\/$//g}"
# use extra command flag on top of the arguments
text = "--with-filename {}"
```

this would allow the user to tweak the value that will get passed to the channel. It could be useful if the channel requires a specific command flag or somewhat different syntax to operate on items

alexpasmantier:
First of all, thanks for yet another great contribution 😊 

I think we should postpone the transitions implementation for the moment, and focus on releasing the current cable redesign without adding any more features for the time being and focusing on refactoring things, fixing bugs, and the major task of updating/writing a whole lot of documentation about the new paradigm as well as a short transition guide so that existing users don't get lost with the upcoming release.

The main reason behind the transitions postponement is that I feel the next step in line after releasing the current cable redesign branch would be to work our way towards implementing actions (global & per-channel) and once that's done I'd like to release it and let it breathe for a while to see how creative we (and the community) can get (things like https://github.com/alexpasmantier/television/issues/352#issuecomment-2888563932 may become possible which could supersede the need for a transitions feature entirely).

lalvarezt:
ok, sounds like a plan 👍. I know, writing new features is more fun than writing docs 😅

lalvarezt:
just pushing this so I don't lose it, it was done since last time anyways

PR #550, opened 2025-06-18
perf: optimization of the hot paths

lalvarezt:
before

```
on   ➜ cargo bench --bench main -- --save-baseline before
    Finished `bench` profile [optimized] target(s) in 0.13s
     Running benches/main.rs (target/release/deps/main-8151fda9d09b6257)
results_list            time:   [16.002 µs 16.095 µs 16.190 µs]
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild

Benchmarking draw: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.6s, or reduce sample count to 80.
draw                    time:   [2.0829 ms 2.2517 ms 2.4277 ms]
```

after

```
on   ➜ cargo bench --bench main -- --baseline before
    Finished `bench` profile [optimized] target(s) in 0.08s
     Running benches/main.rs (target/release/deps/main-8151fda9d09b6257)
results_list            time:   [8.2873 µs 8.3418 µs 8.3963 µs]
                        change: [-49.407% -48.880% -48.360%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) low mild

Benchmarking draw: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.4s, or reduce sample count to 90.
draw                    time:   [1.3345 ms 1.4388 ms 1.5460 ms]
                        change: [-42.872% -36.105% -29.053%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -21,7 +21,10 @@ pub struct Channel {
 
 impl Channel {
     pub fn new(prototype: ChannelPrototype) -> Self {
-        let matcher = Matcher::new(Config::default());
+        let config = Config::default()
+            .prefer_prefix(true)
+            .optimize_for_paths(true);
+        let matcher = Matcher::new(&config);
         let current_source_index = 0;
         Self {
             prototype,
@@ -66,15 +69,19 @@ impl Channel {
 
     pub fn results(&mut self, num_entries: u32, offset: u32) -> Vec<Entry> {
         self.matcher.tick();
-        self.matcher
-            .results(num_entries, offset)
-            .into_iter()
-            .map(|item| {
-                Entry::new(item.inner)
-                    .with_display(item.matched_string)
-                    .with_match_indices(&item.match_indices)
-            })
-            .collect()
+
+        let results = self.matcher.results(num_entries, offset);
+
+        let mut entries = Vec::with_capacity(results.len());
+
+        for item in results {
+            let entry = Entry::new(item.inner)
+                .with_display(item.matched_string)
+                .with_match_indices(&item.match_indices);
+            entries.push(entry);
+        }
+
+        entries
     }
 
     pub fn get_result(&self, index: u32) -> Option<Entry> {
@@ -171,20 +178,22 @@ async fn load_candidates(
         for line in reader.lines() {
             if let Ok(l) = line {
                 if !l.trim().is_empty() {
-                    let () = injector.push(l, |e, cols| {
-                        // PERF: maybe we can avoid cloning here by using &Utf32Str
-                        if let Some(display) = &source.display {
-                            let formatted = display.format(e).unwrap_or_else(|_| {
-                                panic!(
-                                    "Failed to format display expression '{}' with entry '{}'",
-                                    display.raw(),
-                                    e
-                                );
-                            });
-                            cols[0] = formatted.into();
-                        } else {
-                            cols[0] = e.clone().into();
-                        }
+                    // PERF: Optimize string handling - avoid unnecessary clones
+                    let entry_string = if let Some(display) = &source.display {
+                        display.format(&l).unwrap_or_else(|_| {
+                            panic!(
+                                "Failed to format display expression '{}' with entry '{}'",
+                                display.raw(),
+                                l
+                            )
+                        })
+                    } else {
+                        l
+                    };
+
+                    let () = injector.push(entry_string, |e, cols| {
+                        // PERF: Reduced cloning by handling the string more efficiently
+                        cols[0] = e.clone().into();
                     });
                     produced_output = true;
                 }
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -84,25 +84,24 @@ impl Channel {
         entries
     }
 
-    pub fn get_result(&self, index: u32) -> Option<Entry> {
-        self.matcher.get_result(index).map(|item| {
-            let mut entry = Entry::new(item.inner.clone())
-                .with_display(item.matched_string)
-                .with_match_indices(&item.match_indices);
-            if let Some(p) = &self.prototype.preview {
-                // FIXME: this should be done by the previewer instead
-                if let Some(offset_expr) = &p.offset {
-                    let offset_str = offset_expr
-                        .format(&item.inner)
-                        .unwrap_or_else(|_| panic!("Failed to format offset expression '{}' with name '{}'", offset_expr.raw(), item.inner));
+    pub fn get_result(&self, index: u32) -> Entry {
+        let item = self.matcher.get_result(index).expect("Invalid index");
+        let mut entry = Entry::new(item.inner.clone())
+            .with_display(item.matched_string)
+            .with_match_indices(&item.match_indices);
+        if let Some(p) = &self.prototype.preview {
+            // FIXME: this should be done by the previewer instead
+            if let Some(offset_expr) = &p.offset {
+                let offset_str = offset_expr
+                    .format(&item.inner)
+                    .unwrap_or_else(|_| panic!("Failed to format offset expression '{}' with name '{}'", offset_expr.raw(), item.inner));
 
-                    entry = entry.with_line_number(
-                        offset_str.parse::<usize>().unwrap_or(0),
-                    );
-                }
+                entry = entry.with_line_number(
+                    offset_str.parse::<usize>().unwrap_or(0),
+                );
             }
-            entry
-        })
+        }
+        entry
     }
 
     pub fn selected_entries(&self) -> &FxHashSet<Entry> {
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -175,22 +175,20 @@ async fn load_candidates(
         for line in reader.lines() {
             if let Ok(l) = line {
                 if !l.trim().is_empty() {
-                    // PERF: Optimize string handling - avoid unnecessary clones
-                    let entry_string = if let Some(display) = &source.display {
-                        display.format(&l).unwrap_or_else(|_| {
-                            panic!(
-                                "Failed to format display expression '{}' with entry '{}'",
-                                display.raw(),
-                                l
-                            )
-                        })
-                    } else {
-                        l
-                    };
-
-                    let () = injector.push(entry_string, |e, cols| {
-                        // PERF: Reduced cloning by handling the string more efficiently
-                        cols[0] = e.clone().into();
+                    let () = injector.push(l, |e, cols| {
+                        // PERF: maybe we can avoid cloning here by using &Utf32Str
+                        if let Some(display) = &source.display {
+                            let formatted = display.format(e).unwrap_or_else(|_| {
+                                panic!(
+                                    "Failed to format display expression '{}' with entry '{}'",
+                                    display.raw(),
+                                    e
+                                );
+                            });
+                            cols[0] = formatted.into();
+                        } else {
+                            cols[0] = e.clone().into();
+                        }
                     });
                     produced_output = true;
                 }
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -21,9 +21,7 @@ pub struct Channel {
 
 impl Channel {
     pub fn new(prototype: ChannelPrototype) -> Self {
-        let config = Config::default()
-            .prefer_prefix(true)
-            .optimize_for_paths(true);
+        let config = Config::default().prefer_prefix(true);
         let matcher = Matcher::new(&config);
         let current_source_index = 0;
         Self {
--- a/television/matcher/mod.rs
+++ b/television/matcher/mod.rs
@@ -50,7 +50,8 @@ where
     pub status: Status,
     /// The last pattern that was matched against.
     pub last_pattern: String,
-    /// Ephemeral vector to reduce allocations
+    /// A pre-allocated buffer used to collect match indices when fetching the results
+    /// from the matcher. This avoids having to re-allocate on each pass.
     col_indices_buffer: Vec<u32>,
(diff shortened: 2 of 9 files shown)


television/channels/channel.rs
```diff
@@ -21,7 +21,10 @@ pub struct Channel {
 
 impl Channel {
     pub fn new(prototype: ChannelPrototype) -> Self {
-        let matcher = Matcher::new(Config::default());
+        let config = Config::default()
+            .prefer_prefix(true)
+            .optimize_for_paths(true);
```
alexpasmantier: Do we really want this? Entries are not going to be paths most of the time and I'm wondering if doing this would hinder other channels performance.

television/channels/channel.rs
```diff
@@ -171,20 +178,22 @@ async fn load_candidates(
...
+                                l
+                            )
+                        })
+                    } else {
+                        l
+                    };
+
+                    let () = injector.push(entry_string, |e, cols| {
+                        // PERF: Reduced cloning by handling the string more efficiently
+                        cols[0] = e.clone().into();
```
alexpasmantier: I'm not sure what we're gaining here 🤔 
We're still cloning the same string aren't we?

television/channels/channel.rs
alexpasmantier: Ah nevermind, saw the rest of the PR 😅

television/screen/result_item.rs
```diff
@@ -102,32 +103,62 @@ pub fn build_result_line<'a, T: ResultItem + ?Sized>(
...
+        let name_len = entry_name.chars().count();
+        if last_end < name_len {
+            let chars: String = entry_name.chars().skip(last_end).collect();
+            if !chars.is_empty() {
+                spans.push(Span::styled(
+                    chars,
+                    Style::default().fg(colorscheme.result_name_fg),
+                ));
+            }
+        }
```
alexpasmantier: Is this just a stylistic change? I can't seem to spot the difference 🤔

television/utils/strings.rs
```diff
@@ -512,10 +512,32 @@ pub fn make_matched_string_printable(
...
+            if !matched_string
+                .chars()
+                .any(|c| c == '\t' || c == '\n' || c.is_control())
+            {
+                return (matched_string.to_string(), ranges.to_vec());
+            }
+        }
+    }
+
+    // Full processing for non-ASCII strings or strings that need preprocessing
```
alexpasmantier, line 533: I feel like this is what's bringing most of the performance gains along with reusing the same buffer for column indices and pre-allocating for the ui spans.

television/channels/channel.rs
alexpasmantier: Is the idea to then make it configurable on a per-channel basis?

alexpasmantier: I actually believe the option was never plugged in to nucleo so it shouldn't have any effect as is 😅:
https://github.com/lalvarezt/television/blob/f95009486f84c4d174b17cd7d0eacfa483302f6a/television/matcher/config.rs#L61-L67

television/matcher/mod.rs
```diff
@@ -159,32 +162,40 @@ where
...
+                item.matcher_columns[0].slice(..),
+                &mut matcher,
+                &mut self.col_indices_cache,
+            );
+
+            // PERF: Avoid unnecessary sorting if already sorted or small
+            if self.col_indices_cache.len() > 1 {
+                self.col_indices_cache.sort_unstable();
+                self.col_indices_cache.dedup();
+            }
```
alexpasmantier: I'm pretty sure this isn't doing much since the two functions already short-circuit in that case.

television/matcher/mod.rs
```diff
@@ -69,6 +71,7 @@ where
             matched_item_count: 0,
             status: Status::default(),
             last_pattern: String::new(),
+            col_indices_cache: Vec::with_capacity(64), // Pre-allocate for performance
```
alexpasmantier: This is a great optimization but I find the naming quite confusing. When first reading I thought you were caching the indices themselves to avoid having to recompute them later on.

Perhaps `buffer` is a closer term here (`col_indices_buffer`) since it's really just a bunch of memory that we're reusing for multiple purposes (vs a cache from which you'd think you could retrieve information).

television/matcher/mod.rs
```diff
@@ -159,32 +162,40 @@ where
...
+            let matched_string = item.matcher_columns[0].to_string();
+
+            results.push(matched_item::MatchedItem {
+                inner: item.data.clone(),
+                matched_string,
+                match_indices: indices,
+            });
+
+            // Clear for next iteration
+            self.col_indices_cache.clear();
```
alexpasmantier: Not sure we need to `clear` here since we're calling `drain` a couple of lines above.

television/matcher/mod.rs
```diff
@@ -207,12 +218,25 @@ where
...
+            col_indices.dedup();
+
+            let indices = col_indices.drain(..);
             let matched_string = item.matcher_columns[0].to_string();
+
             matched_item::MatchedItem {
                 inner: item.data.clone(),
                 matched_string,
-                match_indices: Vec::new(),
+                match_indices: indices.collect(),
```
alexpasmantier, line 238: I'm actually wondering if we really need these `get_result` methods that fetch a single result since we're already getting the list of results with `results` and the application can then figure out the currently hovered result just using the picker's state 🤔

television/picker.rs
```diff
@@ -86,23 +86,33 @@ impl<T> Picker<T> {
     fn inner_next(&mut self, total_items: usize, height: usize) {
         let selected = self.selected().unwrap_or(0);
         let relative_selected = self.relative_selected().unwrap_or(0);
-        self.select(Some(selected.saturating_add(1) % total_items));
-        self.relative_select(Some((relative_selected + 1).min(height - 1)));
-        if self.selected().unwrap() == 0 {
+
+        let new_selected = selected.wrapping_add(1) % total_items;
```
alexpasmantier, line 90: Nice

television/screen/result_item.rs
```diff
@@ -39,7 +39,8 @@ pub fn build_result_line<'a, T: ResultItem + ?Sized>(
     area_width: u16,
     prefix: Option<bool>, // Some(true)=selected ●, Some(false)=unselected, None=no prefix
 ) -> Line<'a> {
-    let mut spans = Vec::<Span<'a>>::new();
+    // PERF: Pre-allocate spans vector with estimated capacity
+    let mut spans = Vec::<Span<'a>>::with_capacity(8);
```
alexpasmantier: Is 8 a "sweet spot"? Or just intuition?

television/television.rs
```diff
@@ -331,20 +330,33 @@ impl Television {
...
+            for (i, part) in parts.iter().enumerate() {
+                if i > 0 {
+                    result.push(' ');
+                }
+                result.push('\'');
+                result.push_str(part);
+            }
+            result
+        } else {
+            pattern.to_string()
```
alexpasmantier, line 352: I don't think this is worth it (the function isn't called enough to justify the extra complexity) and I'm pretty sure the benefit is negligible compared to your other optimizations.

alexpasmantier:
Questions and nitpicking aside, bravo for that -50% this is great 😁

television/channels/channel.rs
lalvarezt: This is what happens when you push late at night 😅. 

This is not a performance optimization per se, it's a tweak to the matching algorithm to boost paths. But I think we should not do this here, it would be better to put it as a global flag in the cli (like the exact that currently exist)

https://github.com/helix-editor/nucleo/blob/master/matcher/src/config.rs#L49

Let me clean it up and push another commit. If all is good we'll squash them

television/matcher/mod.rs
lalvarezt: it's a micro-optimization, exiting here prevents all the checks and bounds (plus the call and return). Since this is the innermost logic I thought we could shave some nanoseconds per item, it will pay-off if we have 10k+ results. I don't think it hurts anyways

television/channels/channel.rs
lalvarezt: perhaps the comment was misleading, here we have two things
1-we make the closure predictable since we refactored the `if let` logic outside of it. we preprocess the string before the closure 
2-In the else branch: l gets moved, not copied, into entry_string.

television/screen/result_item.rs
lalvarezt: actually there's mistake there on my side, I call multiple time `entry_name.chars()` and that was one of the things I was trying to avoid

television/matcher/mod.rs
lalvarezt: this needs more thought 🤔

television/screen/result_item.rs
lalvarezt: I tried with both 8 and 16 and noticed no difference, just an estimation of how many we could get in a reasonable case

lalvarezt:
what do you know, we squeezed some more

```
on   ➜ cargo bench --bench main -- --baseline before
    Finished `bench` profile [optimized] target(s) in 0.10s
     Running benches/main.rs (target/release/deps/main-8151fda9d09b6257)
results_list            time:   [8.0413 µs 8.1124 µs 8.1839 µs]
                        change: [-50.699% -50.108% -49.534%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

Benchmarking draw: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.4s, or reduce sample count to 90.
draw                    time:   [814.95 µs 852.51 µs 894.11 µs]
                        change: [-65.200% -62.140% -58.281%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

television/picker.rs
alexpasmantier: let's leave that here for the moment, I'll refactor it later in a separate PR

television/matcher/mod.rs
alexpasmantier: let's leave that here for the moment, I'll refactor it later in a separate PR

television/channels/channel.rs
alexpasmantier: I don't think we can do that here.

The reason is we still want to push the raw string (unformatted) into the nucleo matcher as it serves as a basis for other formatting operations such as `output`, `preview` etc which.

```toml
[source]
command = "echo 'foo bar baz'"
display = "first word: {split: :0}"  # uses the raw line outputted by command
output = "second word: {split: :1}"  # uses the raw line outputted by command

[preview]
command = "third word: {split: :2}"  # uses the raw line outputted by command
```

The closure itself is computing what should be matched against from the raw string.

alexpasmantier: i.e. we do want:
```rust
                    let () = injector.push(l, |e, cols| {
                        ...            //  |___ instead of the formatted `entry_string`
```

lalvarezt: my bad, I totally miss that one, I was over confident on the test (that we have to re-check because they didn't fail 🤔). Rolling back that code a leaving the original shows there was not much impact anyways

```
on   ➜ cargo bench --bench main -- --baseline before
    Finished `bench` profile [optimized] target(s) in 0.13s
     Running benches/main.rs (target/release/deps/main-8151fda9d09b6257)
results_list            time:   [8.4621 µs 8.5352 µs 8.6121 µs]
                        change: [-47.697% -47.092% -46.419%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild

Benchmarking draw: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.4s, or reduce sample count to 90.
draw                    time:   [891.20 µs 953.05 µs 1.0219 ms]
                        change: [-62.014% -57.675% -53.028%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
```

alexpasmantier: That's because we don't have any integration tests for the new output, display, etc features 😅

I'll add some this afternoon

television/matcher/mod.rs
```diff
@@ -50,17 +50,19 @@ where
     pub status: Status,
     /// The last pattern that was matched against.
     pub last_pattern: String,
+    /// Ephemeral vector to reduce allocations
```
alexpasmantier:
```suggestion
    /// A pre-allocated buffer used to collect match indices when fetching the results
    /// from the matcher. This avoids having to re-allocate on each pass.
```

television/matcher/mod.rs
```diff
@@ -159,32 +162,38 @@ where
         self.total_item_count = snapshot.item_count();
         self.matched_item_count = snapshot.matched_item_count();
 
-        let mut col_indices = Vec::new();
+        // Clear ephemeral vector for reuse
```
alexpasmantier:
```suggestion
        // Clear the pre-allocated match indices buffer for safety
```

alexpasmantier:
Thanks for the latest changes!

Let's merge this 🎉

PR #551, opened 2025-06-20
perf: pre-allocate result vectors, and reuse picker entries to avoid reallocations

lalvarezt:
before
flamegraph-before

after
flamegraph-after

ran with this command
```
cargo flamegraph --no-inline -c 'record -F 997 --call-graph dwarf --no-inherit' --profile=profiling -- --cable-dir ./cable --config-file ./.config/config.toml files ~/
```

### DIFF
--- a/television/app.rs
+++ b/television/app.rs
@@ -276,7 +276,7 @@ impl App {
             {
                 for event in event_buf.drain(..) {
                     if let Some(action) = self.convert_event_to_action(event) {
-                        if !matches!(action, Action::Tick) {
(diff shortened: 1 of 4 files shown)


alexpasmantier:
For unit enum variants, I'd actually rather use direct comparison which I feel is clearer for the reader.

alexpasmantier:
TLDR: I'd like to keep the direct comparisons in `television/app.rs`, and the rest is great :)

alexpasmantier:
Merging, thanks for the PR 🎉

PR #553, opened 2025-06-20
feat(cli): initial support for source and preview overrides, layout, take_1 and take_1_fast

lalvarezt:


alexpasmantier:
Maybe `source-command` should do more than just overriding the default channel's source command.
When you're doing:
```sh
tv --source-command 'echo 1\\n2\\n3'
```
it feels kind of weird to have all the rest continue to act as if the default channel was playing and try to display a bat preview etc.
image

I guess I would expect that specifying the source-command means I want to bypass the default channel entirely and get a blank canvas:
image

What do you think?

alexpasmantier:
This would probably mean tweaking the `determine_channel` function to build a new prototype on the fly when `source-command` is specified.

lalvarezt:
I totally get your point, we need to make a design decision and document it because it could be jarring when what you expect doesn't happen. For example:

### Use Case 1 (current implementation)

User wants to tweak an existing channel

```bash
tv files --source-command "fd -t f -H"
```

Side-effect is that this next command assumes the default channel if not set

```bash
tv --source-command "fd -t f -H"
```

### Use Case 2 (your proposal)

User wants to create a new channel from scratch

```bash
tv --source-command "fd -t f -H"
# missing channel name and source_command triggers ad-hoc channel
```

Then the concept of default channels disappears because we have a blank canvas that we will expand upon. So all flags will be off and get a minimal interface with default "Custom Channel" input_header.

I guess in this case we could trigger the ad-hoc channel logic only if --source-command is present. Because I don't see much sense of doing that in any other case.

@alexpasmantier What do you think?

alexpasmantier:
Totally on-board with what you said 👍🏻 

- specifying a channel name **and** a `--source-command` should override that channel's source command (`tv files --source-command '...'`)
- not specifying any channel **and** a `--source-command` should start off of a blank prototype (`tv --source-command '...'`)

This makes it possible as you mentioned to create channels on the fly through the cli which I believe is a really nice feature.

lalvarezt:
this should be ready then, unless there's something else you want to add

alexpasmantier (approved):

alexpasmantier:
Very nice!

Thanks for the PR, merging this 🎉

Issue #554, opened 2025-06-21
A `watch`-like feature that periodically reloads the source command


alexpasmantier:
**Feature**
A `watch`-like feature that periodically reloads the source command.

This should be added:
- to the cli: `--watch n` to reload every n seconds
- to channel prototypes: `watch = n` (default 0 which would disable the feature)

**Examples**
This could be useful for custom channels that show updatable state (e.g. a process monitor).

lalvarezt:
@alexpasmantier any thoughts on how to remove? (minimize) the flicker that happens when you manually reload?

alexpasmantier:
We could maybe prevent the transition frame(s) from being rendered while the channel is reloading 🤔

alexpasmantier:
@lalvarezt added a rather simple fix that should do the trick 😊

lalvarezt:
seems to work like a charm, can we use the same logic for channel changes (shortcut and remote)?

alexpasmantier:
Actually not sure what you mean. The fickering while changing channels is just the new channel taking over isn't it?
I can't seem to notice any real flickering on my side 🤔 

EDIT: if you're thinking of applying the same logic to the first few "empty" frames while the channel is loading, I'm afraid that would noticeably degrade the impression of UI responsiveness while changing channels.

EDIT: tried a simple implementation with the same logic and played with it a bit. I feel like the tradeoff isn't worth it. Maybe disabling the default "fillers" like "Select an entry" etc. while changing channels would do?

lalvarezt:
You're probably right, I think it's mostly the perceived change. For example when alternating quickly the preview header shows

t1 -> fileA
t2 -> Select and entry to preview
t3 -> fileB

perhaps we can just avoid that change

EDIT: this would also probably be a sane default (it could be 1000 as well, the point is you don't really need to render 5k lines of preview)
https://github.com/sharkdp/bat?tab=readme-ov-file#fzf

```
[preview]
command = "bat -n --color=always --line-range=:500 {}"
env = { BAT_THEME = "ansi" }
```

alexpasmantier:
Yep see what you mean.

> perhaps we can just avoid that change

That would happen here:
https://github.com/alexpasmantier/television/blob/e517fb800ba315394ea09ac925b145ff770d7311/television/television.rs#L285

So maybe we could avoid resetting the preview state when changing channels and simply rely on the channel itself overwriting the existing `preview_state`?

[...]

PR #557, opened 2025-06-21
feat(cli): add watch flag to trigger reload of channels

lalvarezt:
solves #554

### DIFF
--- a/television/app.rs
+++ b/television/app.rs
@@ -37,6 +37,8 @@ pub struct AppOptions {
     pub preview_size: Option<u16>,
     /// The tick rate of the application in ticks per second (Hz).
     pub tick_rate: f64,
+    /// Watch interval in seconds for automatic reloading (0 = disabled).
+    pub watch_interval: f64,
 }
 
 impl Default for AppOptions {
@@ -51,6 +53,7 @@ impl Default for AppOptions {
             no_preview: false,
             preview_size: Some(DEFAULT_PREVIEW_SIZE),
             tick_rate: default_tick_rate(),
+            watch_interval: 0.0,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


television/app.rs
```diff
@@ -118,6 +123,10 @@ pub struct App {
     /// Render task handle
     render_task: Option<tokio::task::JoinHandle<Result<()>>>,
     options: AppOptions,
+    /// Watch timer task handle for periodic reloading
+    watch_timer_task: Option<tokio::task::JoinHandle<()>>,
+    /// Flag to track if watch mode is active
+    watch_active: bool,
```
alexpasmantier: Could we avoid having this `watch_active` flag by simply having something like:

```rs
impl App {
    fn watch_active(&self) -> bool {
        self.watch_timer_task.is_some()
    }
}
```
?

alexpasmantier: This would rid ourselves of having to manually manage that flag.

television/cli/mod.rs
```diff
@@ -180,18 +182,11 @@ pub fn post_process(cli: Cli) -> PostProcessedCli {
...
-                    layout_str
-                ));
-            }
-        }
-    });
+    let layout: Option<Orientation> =
+        cli.layout.map(|layout_enum| match layout_enum {
+            args::LayoutOrientation::Landscape => Orientation::Landscape,
+            args::LayoutOrientation::Portrait => Orientation::Portrait,
+        });
```
alexpasmantier, line 189: sweet

television/app.rs
lalvarezt: That makes sense 👍

would you mind pushing the change, AFK right now. Otherwise I will change it tonight

alexpasmantier: Yep, done ✅

alexpasmantier:
Works perfectly on my machine, and the implementation is straightforward. Thanks! 🙏🏻 

Merging this 🎉

PR #559, opened 2025-06-22
Keybinding panel

lalvarezt:
Add a new keybinding manager (à la which-key helix style)

it adapts to the current mode

channel mode
image

remote mode
image

Add sorting for remote entries


alexpasmantier:
Very nice!

I'm wondering if we shouldn't drop the top help panel altogether and replace it with this 🤔 That would rid us of a bit of layout complexity too.

lalvarezt:
I was thinking the same thing. But I do like the metadata part, it is useful to know the info there. Perhaps we can keep a streamlined version?

lalvarezt:
FYI I'm working on something already

lalvarezt:
Added status line

image

bubble style
image

alexpasmantier:
Oh wow ahaha that looks so neat!! ❤️ 

Sorry for the late reply on this, got sick. Will take a look now :)

lalvarezt:
> Oh wow ahaha that looks so neat!! ❤️ 
> 
> 
> 
> Sorry for the late reply on this, got sick. Will take a look now :)

Take your time and get well. 
I'm not happy with the theming part, I did minimal effort 😇

I want to reuse the help key binding to raise the status line one or two rows and show the command used and whatever else is useful and completely remove the help menu (sorry for your cool tv icon 🫣). Unless you have something else in mind

alexpasmantier:
> I want to reuse the help key binding to raise the status line one or two rows and show the command used and whatever else is useful and completely remove the help menu (sorry for your cool tv icon 🫣). Unless you have something else in mind

Not sure what you mean, but perhaps the `help: <Ctrl-g> preview: <Ctrl-o>` (or whatever that is set to) could live directly on the status bar as hints instead of being shown on the results panel border?

And in that case, maybe the "which-key" keybindings can be displayed in a horizontal panel that comes in between the status bar and the rest of the UI ?

```
╭─────────────────────────────────────────────────────╮
│                                                     │
│                           UI                        │
│                                                     │
╰─────────────────────────────────────────────────────╯
╭─────────────────────────────────────────────────────╮
│                       keybindings                   │
╰─────────────────────────────────────────────────────╯
------------------------status bar---------------------
```

EDIT: actually since the help panel is getting replaced by the status bar, you can forget about that first `help: <Ctrl-g>` 😊

television/draw.rs
```diff
@@ -135,38 +136,6 @@ impl Ctx {
...
-//         Some(self.instant.cmp(&other.instant))
-//     }
-// }
-//
-// impl Ord for Ctx {
-//     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-//         self.instant.cmp(&other.instant)
-//     }
-// }
-
```
alexpasmantier, line 169: Thanks for cleaning that up 👍🏻

television/screen/keybinding_panel.rs
```diff
@@ -0,0 +1,197 @@
...
+fn generate_keybinding_content(
+    keybindings: &KeyBindings,
+    mode: Mode,
+    colorscheme: &Colorscheme,
+) -> Vec<Line<'static>> {
+    let mut lines = Vec::new();
+
+    // Global keybindings section header
+    lines.push(Line::from(vec![
+        Span::raw(" "), // Left padding
```
alexpasmantier: Ratatui has builtin methods to add padding, should be cleaner than manually adding a space everywhere :p 

https://github.com/lalvarezt/television/blob/989b923e5d86e7a4f024275dc2a033fded94b179/television/screen/help.rs#L77

lalvarezt:
> And in that case, maybe the "which-key" keybindings can be displayed in a horizontal panel that comes in between the status bar and the rest of the UI ?
> 
> ```
> ╭─────────────────────────────────────────────────────╮
> │                                                     │
> │                           UI                        │
> │                                                     │
> ╰─────────────────────────────────────────────────────╯
> ╭─────────────────────────────────────────────────────╮
> │                       keybindings                   │
> ╰─────────────────────────────────────────────────────╯
> ------------------------status bar---------------------
> ```
> 

I propose to leave this one for another PR. If we want to truly make it inline and not floating from the bottom then we need to rearrange by columns to fit in a reasonable size and in a way that looks "nice".

I've implemented the other changes

lalvarezt:
image

image

alexpasmantier:
Added a couple of mods/fixes and a suggestion: shouldn't we simply call this panel "Help"?

lalvarezt:
@alexpasmantier is there anything else you'd like to add to this PR (or to change), I have another one that relies on this one and it's probably better to have them separate

EDIT: not rushing you, is just to have an idea 😅

alexpasmantier:
Nope, just wanted your confirmation on the "help" naming 🙂 
Let's merge this then!

Issue #246, opened 2025-01-07
Packaging for apt-get


alexpasmantier:
**Problem**
Tv is not currently installable via `apt-get`

**Feature**
```bash
sudo apt-get install television
```

kapobajza:
Hey @alexpasmantier. I've taken a look into this issue and I've got an idea on how it can be done. We basically have 2 options:
1. Host our own apt repository (through GH pages, for example)
2. Submit the binary to the official Debian repository. This can take quite a lot of time, as far as I've researched

So we could host our own apt repository and submit the binary to the official repository simultaneously. And then add an installation method of the former approach to our docs, until the binary is accepted and available through the official repo. Once it's accepted, update the docs to simply run `sudo apt-get install televesion` instead of adding custom repostories prior to running the `apt-get install` command.

What do you think about this?

I've created an example repo where I hosted a small program that simply prints "Hello world" to the console, just to test this out. You can find the repo [here](https://github.com/kapobajza/apt-hosting-test). The readme of that repo is basically how the install procedure for the `television` program would look like until the binary is accepted by the official Debian repository.

alexpasmantier:
Nice job on the investigation!

I feel like 1. is probably quite a hassle until we have proper gh pages for the project and would involve quite some CI code to keep up to date (but maybe I'm mistaken). 

The current way to install on most apt-based distributions is https://github.com/alexpasmantier/television/wiki/Installation#--debian-based-debian-ubuntu-pop_os-linux-mint-etc which isn't great tbh but could become more tolerable if we're waiting on 2.

Do you have any idea of what 2. entails / what the typical time-window is / how we might speed it up / etc?

We could still go with 1. in the meantime if you feel motivated, but it's unclear to me at this stage if it would really be worth the ride.

kapobajza:
Hey @alexpasmantier, sorry for the delayed reply.

I thought that the project actually already has GH pages set up, so that's why I mentioned it. And yes, approach 1. would require CI code: to create packaging info about the `deb` package, create hash checksums, sign the package, etc.

> The current way to install on most apt-based distributions is https://github.com/alexpasmantier/television/wiki/Installation#--debian-based-debian-ubuntu-pop_os-linux-mint-etc which isn't great tbh but could become more tolerable if we're waiting on 2.

[...]

PR #568, opened 2025-06-28
feat(binary): Host our own apt repo

kapobajza:
This PR solves part of https://github.com/alexpasmantier/television/issues/246, since hosting our own apt repo is just the first step.

Right now it's a draft, because I put the CD code inside of the [.github/workflows/static.yml](https://github.com/alexpasmantier/television/blob/98a3d7b8547268084e5f54f48eb12774e6969d36/.github/workflows/static.yml) file and this, most-likely, doesn't belong there. Where exactly to put it will need to be discussed, but the requirements are:
- to start executing this code after we generate the `.deb` binary
- and to deploy the static GH pages content after we've executed this code

I've deployed a test version of the `.deb` package to my GH pages and you can actually already install `television` via `apt` like this:

```sh
# download the repository public key and put it in the trusted directory
curl -sS https://kapobajza.github.io/television/apt/repository-key.asc | sudo gpg --dearmor --yes -o /etc/apt/trusted.gpg.d/television.gpg

# put the repository download link inside the apt sources list
echo "deb https://kapobajza.github.io/television/apt stable main" | sudo tee /etc/apt/sources.list.d/television.list

sudo apt update

# install the package 🚀 
sudo apt install television
```


alexpasmantier:
Great work!

I'm thinking this can go inside of `.github/workflows/cd.yml` for now as a new job. That way it always gets run after the new release is published.

```yml
jobs:
  create-release:
    ...
  publish-release:
    ...
  packaging:
    [here]
```

What do you think?

kapobajza:
> Great work!
> 
> I'm thinking this can go inside of `.github/workflows/cd.yml` for now as a new job. That way it always gets run after the new release is published.
> 
> ```yaml
> jobs:
>   create-release:
>     ...
>   publish-release:
>     ...
>   packaging:
>     [here]
> ```
> 
> What do you think?

Yeah, makes sense. I can put it there. Should I also put the GH pages deployment steps there? After the packaging is done, we need to deploy it to GH pages.


Btw for this to work you would also need to generate a GPG key and put it inside the repo secrets, it's being used [for this action](https://github.com/alexpasmantier/television/blob/6f7ccea3a105694d29a0eab8231b7828f2a46da1/.github/workflows/static.yml#L37-L38). And we'll also need to update the key id that's being used [here](https://github.com/alexpasmantier/television/blob/2a15a799cf657c8ea8337b177ebdb48f657437c4/apt/reprepro/conf/distributions#L5). Instructions on how to generate a GPG key are:
1. Run `gpg --full-generate-key`
2. Select RSA (sign only) with a key size of 4096 bits, and follow the prompts to create your key
3. Run `gpg --armor --export-secret-key email@example.com` - the email is the one you entered when generating your key in step 1. This will print out your GPG private key ([`secrets.GPG_PRIVATE_KEY`](https://github.com/alexpasmantier/television/blob/6f7ccea3a105694d29a0eab8231b7828f2a46da1/.github/workflows/static.yml#L37))
4. [`secrets.GPG_PASSPHRASE`](https://github.com/alexpasmantier/television/blob/6f7ccea3a105694d29a0eab8231b7828f2a46da1/.github/workflows/static.yml#L38) is, obviously, the passphrase you entered when generating the key
5. Run `gpg --list-keys --keyid-format long`. The id after `rsa4096/...` is your key id that should be used to [replace the existing one](https://github.com/alexpasmantier/television/blob/2a15a799cf657c8ea8337b177ebdb48f657437c4/apt/reprepro/conf/distributions#L5). For example in my case it's:

```
# The key id is 43BD55100555DBF2
pub   rsa4096/43BD55100555DBF2 2025-06-25 [SC]
      199782ABD367B59116D4447E43BD55100555DBF2
```

And that's it.

Sorry for the longer explanation, you are probably aware or may have done something like this in the past. But in case you didn't, I provided a guide with details. 😅

alexpasmantier:
I feel the way to go should be to have the GH pages deployment in a separate workflow which would be triggered based on certain events, typically once the CD workflow finishes:
```yml
name: Deploy to GH pages
on:
  workflow_run:
    workflows: ["CD"]
    branches: [main]
    types: 
      - completed

jobs:
  ...
```

alexpasmantier:
@kapobajza Rebased the PR on main, added the GPG public keys to the different files and added the secrets to the repo 🎉 

EDIT: the rebase makes it look like I'm the author of the commits but that's just github being github.
image

kapobajza:
> @kapobajza Rebased the PR on main, added the GPG public keys to the different files and added the secrets to the repo 🎉
> 
> EDIT: the rebase makes it look like I'm the author of the commits but that's just github being github. image

Great, thank you for that. 😄

I've moved all the ci/cd code from `static.yml` to `cd.yml`. So the PR is no longer a draft.
Just beware that we won't be able to test this out properly until we make a new release, obviously 😅. Since the `.deb` files that were released previously didn't contain the [additional needed metadata](https://github.com/alexpasmantier/television/blob/0580e1fb1758b2d626f6e0015946c4b4bc3a2f1a/Cargo.toml#L139-L151), [this `reprerpo` command](https://github.com/alexpasmantier/television/blob/6e33212923efc23b5f3b29665718a195d2409916/.github/workflows/cd.yml#L340) will fail.

And I have a question: I wanted to include the packaging for `aarch64` also, but I've noticed that the [target is commented out](https://github.com/alexpasmantier/television/blob/6e33212923efc23b5f3b29665718a195d2409916/.github/workflows/cd.yml#L203-L205). Why is that?

alexpasmantier:
> Just beware that we won't be able to test this out properly until we make a new release, obviously 😅. Since the .deb files that were released previously didn't contain the [additional needed metadata](https://github.com/alexpasmantier/television/blob/0580e1fb1758b2d626f6e0015946c4b4bc3a2f1a/Cargo.toml#L139-L151), [this reprerpo command](https://github.com/alexpasmantier/television/blob/6e33212923efc23b5f3b29665718a195d2409916/.github/workflows/cd.yml#L340) will fail.

No problem, we'll see how it goes when we release 0.12 and will iterate from there 👍🏻 

> And I have a question: I wanted to include the packaging for aarch64 also, but I've noticed that the [target is commented out](https://github.com/alexpasmantier/television/blob/6e33212923efc23b5f3b29665718a195d2409916/.github/workflows/cd.yml#L203-L205). Why is that?

I believe I commented that out because compilation would fail on aarch64 because of a C library I was relying on for builtin syntax highlighting (namely oniguruma). But this isn't a dependency anymore so it should be fine to re-add that line in the CD workflow 👍🏻

alexpasmantier:
@kapobajza should we merge this or do you want to wait for the changelog thing to be done?

kapobajza:
> @kapobajza should we merge this or do you want to wait for the changelog thing to be done?

I've just added the deb packaging for the `aarch64` architecture. The testing I could do is very limited, so I would need to do proper testing once we release the 0.12 version.

But if everything's fine from your side, we can go ahead and merge this. We don't have to wait for the changelog thing, that could be a separate PR.

PR #569, opened 2025-06-28
fix(cli): fix parsing of arguments for autocomplete-prompt

lalvarezt:
the new clap restrictions introduced a regression with `--autocomplete-prompt` causing the following command to fail

```
tv --autocomplete-prompt "nvim" /etc
```

it incorrectly interpreted that `/etc/` was a channel and triggered the channel|autocomplete-prompt restriction

### DIFF
--- a/tests/cli/cli_special.rs
+++ b/tests/cli/cli_special.rs
@@ -43,6 +43,24 @@ fn test_autocomplete_prompt_and_channel_argument_conflict_errors() {
     tester.assert_raw_output_contains("cannot be used with");
 }
 
+/// Tests that --autocomplete-prompt works with a working directory path argument.
+#[test]
+fn test_autocomplete_prompt_with_working_directory() {
+    let mut tester = PtyTester::new();
+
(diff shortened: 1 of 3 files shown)


tests/cli/cli_special.rs
```diff
@@ -43,6 +43,24 @@ fn test_autocomplete_prompt_and_channel_argument_conflict_errors() {
...
+        "ls",
+        "/etc",
+    ]);
+    let mut child = tester.spawn_command_tui(cmd);
+
+    // Send Ctrl+C to exit (the test is mainly to ensure no CLI parsing error)
+    tester.send(&ctrl('c'));
+    PtyTester::assert_exit_ok(&mut child, DEFAULT_DELAY);
+}
+
```
alexpasmantier, line 63: nice 👍🏻

alexpasmantier:
Thanks for the fix 🙏🏻

alexpasmantier (approved):

PR #571, opened 2025-06-29
fix(keybindings): add cmd as an option for modifrs

domaschh:
Adds the cmd modifier key

### DIFF
--- a/television/config/keybindings.rs
+++ b/television/config/keybindings.rs
@@ -97,21 +97,27 @@ fn extract_modifiers(raw: &str) -> (&str, KeyModifiers) {
     let mut current = raw;
 
     loop {
-        match current {
-            rest if rest.starts_with("ctrl-") => {
-                modifiers.insert(KeyModifiers::CONTROL);
-                current = &rest[5..];
-            }
-            rest if rest.starts_with("alt-") => {
-                modifiers.insert(KeyModifiers::ALT);
-                current = &rest[4..];
-            }
-            rest if rest.starts_with("shift-") => {
-                modifiers.insert(KeyModifiers::SHIFT);
-                current = &rest[6..];
-            }
-            _ => break, // break out of the loop if no known prefix is detected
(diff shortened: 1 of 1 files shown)


television/config/keybindings.rs
```diff
@@ -97,21 +97,27 @@ fn extract_modifiers(raw: &str) -> (&str, KeyModifiers) {
...
+        }
+        if let Some(rest) = current.strip_prefix("shift-") {
+            modifiers.insert(KeyModifiers::SHIFT);
+            current = rest;
+            continue;
+        }
+        if let Some(rest) = current.strip_prefix("alt-") {
+            modifiers.insert(KeyModifiers::ALT);
+            current = rest;
+            continue;
```
alexpasmantier, line 113: nicer indeed 👍🏻

television/config/keybindings.rs
```diff
@@ -97,21 +97,27 @@ fn extract_modifiers(raw: &str) -> (&str, KeyModifiers) {
...
         }
+        if let Some(rest) = current.strip_prefix("cmd-") {
+            modifiers.insert(KeyModifiers::SUPER);
+            current = rest;
+            continue;
+        }
+        break;
```
alexpasmantier: Should this maybe be dependent on the target os and be `cmd-` for macOS and `super-` for linux?

domaschh: I mean makes sense I'm just coming from the zed editor there cmd is used for both. I don't have any preference tho.

television/config/keybindings.rs
domaschh: ❤️

television/config/keybindings.rs
alexpasmantier: Added the change so that users may also use `super-` 👍🏻

alexpasmantier:
Thanks so much for the PR 🙏🏻 

Merging this 🎉

PR #573, opened 2025-06-29
feat: add global/channel input history

lalvarezt:
sometimes I want to tweak the previous search, or just run it again. this solves that problem


.config/config.toml
```diff
@@ -18,6 +18,18 @@
 tick_rate = 50
 default_channel = "files"
 
+# History settings
+# ---------------
+# Maximum number of entries to keep in the global history (default: 100)
+# The history tracks search queries across all channels and sessions
+# Set to 0 to disable history functionality entirely
+history_size = 100
```
alexpasmantier: does the channel always take precedence on this?

alexpasmantier:
Cool feature! Tested it out and it mostly seems to work nicely 👍🏻

.config/config.toml
```diff
@@ -18,6 +18,18 @@
...
+
+# Whether to use global history mode (default: false)
+# When true: history navigation shows entries from all channels
+# When false: history navigation is scoped to the current channel
+global_history_mode = false
```
alexpasmantier:
```suggestion
global_history = false
```

television/app.rs
```diff
@@ -278,6 +298,39 @@ impl App {
...
+                        "Failed to create history for channel {}: {}",
+                        channel_prototype.metadata.name, e
+                    );
+                    self.history = None;
+                }
+            }
+        } else {
+            // History disabled for this channel
+            self.history = None;
+        }
```
alexpasmantier: Couldn't you get rid of this function and do something like:
```rs
self.history.channel_name = &self.television.channel_prototype.clone();
```

Since you're using one single file for everything there shouldn't be any need to go reading from disk when switching channels. You just want to update the channel name so that the channel-specific navigation still works.

television/app.rs
```diff
@@ -570,10 +655,12 @@ impl App {
                     && self.television.mode == Mode::Channel
                 {
                     self.update_keymap();
+                    self.update_history();
```
alexpasmantier, line 633: couldn't this be:
```rs
self.history.channel_name = &self.television.channel_prototype.clone();
```
?

television/cli/args.rs
```diff
@@ -403,6 +403,15 @@ pub struct Cli {
...
     pub cable_dir: Option<String>,
 
+    /// Use global history mode instead of channel-specific history.
+    ///
+    /// This flag only works in channel mode.
+    ///
+    /// When enabled, history navigation will show entries from all channels.
+    /// When disabled (default), history navigation is scoped to the current channel.
+    #[arg(long, verbatim_doc_comment)]
+    pub global_history_mode: bool,
```
alexpasmantier:
```suggestion
    pub global_history: bool,
```

television/cli/mod.rs
```diff
@@ -96,6 +96,9 @@ pub struct PostProcessedCli {
     pub tick_rate: Option<f64>,
     pub watch_interval: Option<f64>,
 
+    // History configuration
+    pub global_history_mode: bool,
```
alexpasmantier:
```suggestion
    pub global_history: bool,
```

television/history.rs
```diff
@@ -0,0 +1,275 @@
...
+use tracing::error;
+
+const HISTORY_FILE_NAME: &str = "history.json";
+pub const DEFAULT_HISTORY_SIZE: usize = 100;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct HistoryEntry {
+    pub entry: String, // The search query/pattern that was typed
+    pub channel: String,
+    pub timestamp: u64,
```
alexpasmantier:
```suggestion
    /// The search query/pattern that was typed
    pub entry: String,
    pub channel: String,
    pub timestamp: u64,
```

television/history.rs
```diff
@@ -0,0 +1,275 @@
...
+            current_index: None,
+            max_size,
+            file_path,
+            current_channel: channel_name.to_string(),
+            global_mode,
+        };
+
+        history.load_from_file()?;
+        Ok(history)
+    }
```
alexpasmantier: I think the constructor shouldn't have any side effects and should return `Self` and not call `history.load_from_file()`.
You would then call an init method that has the side effect:
```rs
let history = History::new(...);  // or History::from(&config)
history.init()?;
```

This makes testing much easier since you can avoid the side effect if you don't want fs interaction and solely work in memory.

television/history.rs
```diff
@@ -0,0 +1,275 @@
...
+
+        // Reset current index when adding new entry
+        self.current_index = None;
+
+        // Trim history if it exceeds max size
+        if self.entries.len() > self.max_size {
+            self.entries.drain(0..self.entries.len() - self.max_size);
+        }
+
+        self.save_to_file()?;
```
alexpasmantier: same here, this shouldn't save to file but just update the struct in-memory.

television/config/mod.rs
```diff
@@ -40,17 +41,33 @@ pub struct AppConfig {
     /// The default channel to use when no channel is specified
     #[serde(default = "default_channel")]
     pub default_channel: String,
+    /// Maximum number of entries to keep in the global history
+    #[serde(default = "default_history_size")]
+    pub history_size: usize,
+    /// Whether to use global history mode (all channels) or channel-specific history (default)
+    #[serde(default = "default_global_history_mode")]
+    pub global_history_mode: bool,
```
alexpasmantier:
```suggestion
    pub global_history: bool,
```

alexpasmantier:
This adds quite a bit of new logic and for a feature we never really talked about (in a design sense). I think the best way to move forward would be to have a specification of what all this should be doing in the form of tests.

television/history.rs
```diff
@@ -0,0 +1,275 @@
...
+            if entry.channel == channel {
+                let idx = search_start + offset;
+                self.current_index = Some(idx);
+                return Some(entry);
+            }
+        }
+
+        // Reached the end, reset navigation
+        self.current_index = None;
+        None
```
alexpasmantier: couldn't this use `get_next` and filter on entries that have the corresponding channel?

television/history.rs
```diff
@@ -0,0 +1,275 @@
...
+            self.entries[..search_start].iter().enumerate().rev()
+        {
+            if entry.channel == channel {
+                self.current_index = Some(idx);
+                return Some(entry);
+            }
+        }
+
+        None
+    }
```
alexpasmantier: couldn't this use `get_previous` and filter on entries that have the corresponding channel?

television/history.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    /// Get the next history entry based on the configured mode.
+    pub fn get_next_entry(&mut self) -> Option<&HistoryEntry> {
+        if self.global_mode {
+            self.get_next()
+        } else {
+            let channel = self.current_channel.clone();
+            self.get_next_in_channel(&channel)
+        }
+    }
```
alexpasmantier: it feels like these should be the only public getters. That would avoid confusion on which one to use between `get_next_entry`, `get_next`, `get_next_for_channel` etc.

lalvarezt:
> This adds quite a bit of new logic and for a feature we never really talked about (in a design sense). I think the best way to move forward would be to have a specification of what all this should be doing in the form of tests.

sure, let me make some changes and iterate on it
thanks for the detailed comments

lalvarezt:
you've raised a valid point, what would be your preferred way to approach this? Should I first create an item in discussions to go over potential new features?

alexpasmantier:
> you've raised a valid point, what would be your preferred way to approach this? Should I first create an item in discussions to go over potential new features?

For the next ones.
But for this one I think we just need a spec of what this does and how it should behave. This spec can take the form of tests so that people building on it afterwards know if they're breaking anything when making changes.

EDIT: and maybe not discussions but rather creating an issue with the `enhancement` label and the new feature template.

lalvarezt:
added reasonable number of tests, this should be the base to analyze some of the changes you proposed

.config/config.toml
lalvarezt: fixed it, I was ignoring the config.toml value

television/app.rs
lalvarezt: good catch, fixed

television/history.rs
lalvarezt: this is the only place where it is saved, and this is called when you select an entry and exit the app

EDIT: you mean you'd rather save here `Action::SelectAndExit` instead?

alexpasmantier: Yes, `add_entry` should really just add an entry to in the in-memory struct.
And persisting should be done by calling the separate method `save_to_file` where desired.
So yeah, calling `save_to_file` from where you want to write to disk in the calling code 👍🏻

lalvarezt:
I think I covered all of your original concerns. please have a look again

docs/cli.md
```diff
@@ -435,6 +435,17 @@ Television's options are organized by functionality. Each option behaves differe
 - **Default**: `~/.config/tv/cable/` (Linux/macOS) or `%APPDATA%\tv\cable\` (Windows)
 - **Use Case**: Custom channel collections or shared team channels
 
+### 📚 History Options
+
+#### `--global-history-mode`
```
alexpasmantier:
```suggestion
#### `--global-history`
```

tests/cli/cli_history.rs
alexpasmantier, line 1: Because these tests mainly test the logic of `history.rs` (by contrast to testing the entire application), let's put them directly inside `television/history.rs` as unit tests.

domaschh:
Hi what's the status on this any way I can help?:)

alexpasmantier:
Took the liberty of dropping the `max_size` per channel feature as I felt it brought more confusion to users than anything else and burdened up the code by being an `Option<usize>`.

This means setting `max_size` to 0 in a user's config file will disable **persisting** and **loading** the history to disk but the user will still have access to the feature within a given session (if we would ever want to register queries when a user executes actions other than selecting and exiting). This does mean you always carry around the `History` struct but the cost is really negligible.

Other than that I'm happy to merge this if you're ready to proceed 😊

lalvarezt:
> Took the liberty of dropping the `max_size` per channel feature as I felt it brought more confusion to users than anything else and burdened up the code by being an `Option<usize>`.
> 
> This means setting `max_size` to 0 in a user's config file will disable **persisting** and **loading** the history to disk but the user will still have access to the feature within a given session (if we would ever want to register queries when a user executes actions other than selecting and exiting). This does mean you always carry around the `History` struct but the cost is really negligible.
> 
> Other than that I'm happy to merge this if you're ready to proceed 😊

sure thing, looks good to me. The costs are not that big considering the simplification of the code
thanks for taking the time and do the changes

alexpasmantier:
Merging this 🎉

Issue #575, opened 2025-06-30
Add support for non-fullscree, i.e height limited overlay


lalvarezt:
**Problem**
Current the TUI takes up the whole screen, it could be handy to have a partial view that's less intrusive

**Feature**
A cli flag could be used to the set the height of the window

**Examples**
It would look like this

![Image](https://github.com/user-attachments/assets/e5ed6f60-4561-4be9-a30c-4bf0ac07271a)

alexpasmantier:
That does sound like a great feature, especially for shell integration 👍

But you probably would have to not use the alternate screen and that will come with its set of constraints (cleaning up properly etc).

Do you have any idea of how you want to approach this yet?

lalvarezt:
yes, I have a working proto that does something like

in `render.rs`
```rust
                RenderingTask::Render(context) => {
                    if let Ok(size) = tui.size() {
                        // Ratatui uses `u16`s to encode terminal dimensions and its
                        // content for each terminal cell is stored linearly in a
                        // buffer with a `u16` index which means we can't support
                        // terminal areas larger than `u16::MAX`.
                        if size.width.checked_mul(size.height).is_some() {
                            queue!(stderr(), BeginSynchronizedUpdate).ok();
                            let overlay_row = tui.base_row();
                            tui.terminal.draw(|frame| {
                                let drawing_area = if let Some(h) = height {
                                    let terminal_area = frame.area();
                                    let ui_height =
                                        h.min(terminal_area.height);
                                    Rect {
                                        x: 0,
                                        y: overlay_row,
                                        width: terminal_area.width,
                                        height: ui_height,
                                    }
                                } else {
                                    frame.area()
                                };

                                match draw(&context, frame, drawing_area) {
                                    Ok(layout) => {
                                        if layout != context.layout {
...
```

and in `tui.rs`

```rust
    pub fn enter(&mut self) -> Result<()> {
        enable_raw_mode()?;
        let mut buffered_stderr = LineWriter::new(stderr());

        if self.fullscreen {
            execute!(buffered_stderr, EnterAlternateScreen)?;
            execute!(buffered_stderr, EnableMouseCapture)?;
            self.terminal.clear()?;
        } else {
            // print <height> new-lines on stdout – this may cause scroll.
            let h = self.height.unwrap_or(10);
            {
                let mut out: StdoutLock<'_> = stdout().lock();
                for _ in 0..h {
                    writeln!(out)?;
                }
                out.flush()?;
            }

[...]

PR #578, opened 2025-07-01
Adds support for non-fullscreen interfaces

lalvarezt:
implements #575

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -346,6 +346,15 @@ Television's options are organized by functionality. Each option behaves differe
 - **Conflicts**: Cannot be used with `--inline`
 - **Use Case**: Precise control over interface height
 
+#### `--width <INTEGER>`
+
+**Purpose**: Sets a fixed width for non-fullscreen mode
+
+- **Both Modes**: Same behavior
+- **Range**: 10 or higher (minimum UI width required)
+- **Requires**: Must be used with `--inline` or `--height`
+- **Use Case**: Precise control over interface width
+
 #### `--inline`
 
 **Purpose**: Uses all available empty space at the bottom of the terminal
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -337,6 +337,24 @@ Television's options are organized by functionality. Each option behaves differe
 - **Range**: 10-100%
 - **Use Case**: Adapt to different screen sizes or preferences
 
+#### `--height <INTEGER>`
+
+**Purpose**: Sets a fixed height for non-fullscreen mode
+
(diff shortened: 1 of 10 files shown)


lalvarezt:
possible next steps 

- [x] it seems like it would be rather straightforward to add configurable width.
- [x] don't put any max value for the height or the width in the cli (clap value parsers) since we're already guarding against that when rendering by taking the min of that value and the actual available space
- [x] we probably want a default non-fullscreen mode that just takes up what's available (e.g. tv --inline)

lalvarezt:
@alexpasmantier finished, can you have a look

alexpasmantier:
Great! This looks awesome, congrats.

Will be very useful, especially for shell integration and other applications like that 👍🏻 

Would you mind adding a couple of very simple snapshot-based PTY tests that check the UI is rendering to the specified size?

lalvarezt:
My thoughts exactly. This is one of the remaining things I had on FZF, after is done I can fully migrate

I had a look at the testing part and is not that simple. The current PTYtester doesn't capture "overlay" mode (maybe because of alternate screen?). I'll have a look later tonight to see if I can make it work

alexpasmantier:
> I had a look at the testing part and is not that simple. The current PTYtester doesn't capture "overlay" mode (maybe because of alternate screen?). I'll have a look later tonight to see if I can make it work

Ah see what you mean. We can leave that aside for now if it's too much of a hassle 👍🏻 

Shouldn't we be able to use `--width` without specifying a `--height` ? (which would take all the height available)
```
television partial-view*​ ≡
❯ cargo run -- --input 'television' --preview-size 70 files --width 60
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/tv --input television --preview-size 70 files --width 60`
Error parsing CLI arguments: --width can only be used in combination with --inline or --height
```

lalvarezt:
> Shouldn't we be able to use `--width` without specifying a `--height` ? (which would take all the height available)

No, you can do either
--inline --width 70
Or
--height 20 --width 70

--width only makes no sense because by default we're in full screen mode

alexpasmantier:
There are a couple of conflicts with main which should be solved by rebasing the branch.

As I said earlier, I'm happy to merge this as is, except if you've figured out a way of testing the layout 😊

lalvarezt:
I say we close it for now and deal with the testing part later

alexpasmantier:
Merged 🎉 

Thanks for the awesome feature 🙏🏻

PR #584, opened 2025-07-02
Minor changes

lalvarezt:
- Improved determine channel logic, it will make easier to maintain moving forward
- Fixed hint logic for preview, it had issues for stdin and ad-hoc channels
- Override logic for CLI UI elements was not handled consistently
- Added few more tests

### DIFF
--- a/television/main.rs
+++ b/television/main.rs
@@ -1,6 +1,5 @@
 use anyhow::Result;
 use clap::Parser;
-use rustc_hash::FxHashMap;
 use std::env;
 use std::io::{BufWriter, IsTerminal, Write, stdout};
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


television/main.rs
```diff
@@ -251,26 +251,209 @@ pub fn handle_subcommand(command: &Command, config: &Config) -> Result<()> {
     }
 }
 
+/// Creates a `CommandSpec` from a `Template` with default settings
+fn create_command_spec(template: Template) -> CommandSpec {
+    CommandSpec::new(vec![template], false, FxHashMap::default())
+}
```
alexpasmantier: This could directly be `CommandSpec::from::<Template>`

alexpasmantier:
This is nicer indeed, thanks for taking care of that 😊

lalvarezt:
made some QoL changes as suggested, if you're ok with it we can merge it

PR #585, opened 2025-07-02
feat(tui): add special testing conditions for overlay testing

lalvarezt:
the issue was related to the starting row. And this is the same reason we cannot test --inline, in the test env there's no way to calculate the amount of empty lines (or I don't see it). But at least --height and --width are testable


alexpasmantier:
Good job on figuring that out.

The env var logic is kind of annoying but does solve the problem. Let's keep that for now 👍🏻

Issue #586, opened 2025-07-03
The new inline mode --inline or --height don't work for shell integration


lalvarezt:
**Description**
Height limited interfaces don't work correctly during shell integration

**Example**
run

```sh
# works as expected in full-screen mode
echo $(tv) 

# doesn't work
echo $(tv --height 20)
echo $(tv --inline)
```

**Expected behavior**
Show a height bound interface in the same way as running with tty

**Actual behavior**

Get one of these two errors

Error: The cursor position could not be read within a normal duration
Error: channel closed

**Environment**
 - Nothing specific

**Additional Context**
Atuin somehow solves this problem, they also use `ratatui` and `crossterm` so there must be something we could do to mimic it

lalvarezt:
finally solved this, but we will have to do a funny merge later on with the changes from #587

lalvarezt:
leaving this unmerged PR from `crossterm` here for later, it is related
https://github.com/crossterm-rs/crossterm/pull/957
@alexpasmantier the discussion over there is interesting

alexpasmantier:
Fixed by #588

raylu:
I just installed television 0.12.0 and the default ctrl+r bind doesn't work when there isn't enough space at the bottom of my terminal

after clearing my screen:
<img width="417" height="397" alt="Image" src="https://github.com/user-attachments/assets/f928cde6-06dd-4f68-a4d3-43bdee20bca9" />
with no space at the bottom of my terminal:
<img width="792" height="535" alt="Image" src="https://github.com/user-attachments/assets/af6f5aa1-9155-439f-ad1d-7247154d9409" />
you can see it tried to make space but it's not using it
invoking `tv --inline` has this problem, but `tv --height 20` does not
resizing the terminal while `--inline` is running fixes the issue

in general, I'll say I'm pretty confused about what `--inline` does. it says
>  If there is insufficient space to meet the minimum height the terminal will scroll.

but what is the "minimum height"? you can't use `--height` with `--inline`

---

I'm on windows terminal 1.22.111410.0
```
$ zsh --version
zsh 5.9 (x86_64-debian-linux-gnu)
$ tv --version
television 0.12.0
```

PR #588, opened 2025-07-03
fix(tui): fixed shell completion widget not rendering, add poc for fish

lalvarezt:
solves #586

### DIFF
--- a/television/channels/entry.rs
+++ b/television/channels/entry.rs
@@ -126,7 +126,7 @@ impl Entry {
         self.display.as_deref().unwrap_or(&self.raw)
     }
 
-    pub fn stdout_repr(&self, template: &Option<Template>) -> String {
+    pub fn std_repr(&self, template: &Option<Template>) -> String {
         if let Some(template) = template {
             return template.format(&self.raw).unwrap_or_else(|_| {
                 panic!(
@@ -195,7 +195,7 @@ mod tests {
             line_number: None,
         };
         assert_eq!(
-            entry.stdout_repr(&Some(Template::parse("{}").unwrap())),
+            entry.std_repr(&Some(Template::parse("{}").unwrap())),
             "test name with spaces"
         );
     }
--- a/television/channels/entry.rs
+++ b/television/channels/entry.rs
@@ -126,7 +126,7 @@ impl Entry {
         self.display.as_deref().unwrap_or(&self.raw)
     }
 
-    pub fn std_repr(&self, template: &Option<Template>) -> String {
+    pub fn output(&self, template: &Option<Template>) -> String {
         if let Some(template) = template {
             return template.format(&self.raw).unwrap_or_else(|_| {
                 panic!(
@@ -195,7 +195,7 @@ mod tests {
             line_number: None,
         };
         assert_eq!(
-            entry.std_repr(&Some(Template::parse("{}").unwrap())),
+            entry.output(&Some(Template::parse("{}").unwrap())),
             "test name with spaces"
         );
     }
--- a/television/main.rs
+++ b/television/main.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use clap::Parser;
 use std::env;
-use std::io::{BufWriter, IsTerminal, Write, stderr, stdout};
+use std::io::{BufWriter, IsTerminal, Write, stdout};
 use std::path::PathBuf;
 use std::process::exit;
 use television::{
@@ -120,8 +120,8 @@ async fn main() -> Result<()> {
     debug!("Running application...");
     let output = app.run(stdout().is_terminal(), false).await?;
     info!("App output: {:?}", output);
-    let stderr_handle = stderr().lock();
-    let mut bufwriter = BufWriter::new(stderr_handle);
+    let stdout_handle = stdout().lock();
+    let mut bufwriter = BufWriter::new(stdout_handle);
     if let Some(entries) = output.selected_entries {
         for entry in &entries {
             writeln!(
--- a/television/main.rs
+++ b/television/main.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use clap::Parser;
 use std::env;
-use std::io::{BufWriter, IsTerminal, Write, stdout};
+use std::io::{BufWriter, IsTerminal, Write, stderr, stdout};
 use std::path::PathBuf;
 use std::process::exit;
 use television::{
@@ -120,16 +120,15 @@ async fn main() -> Result<()> {
     debug!("Running application...");
     let output = app.run(stdout().is_terminal(), false).await?;
     info!("App output: {:?}", output);
-    let stdout_handle = stdout().lock();
-    let mut bufwriter = BufWriter::new(stdout_handle);
+    let stderr_handle = stderr().lock();
+    let mut bufwriter = BufWriter::new(stderr_handle);
     if let Some(entries) = output.selected_entries {
         for entry in &entries {
             writeln!(
                 bufwriter,
                 "{}",
-                entry.stdout_repr(
-                    &app.television.channel.prototype.source.output
-                )
+                entry
+                    .std_repr(&app.television.channel.prototype.source.output)
             )?;
         }
     }
--- a/television/main.rs
+++ b/television/main.rs
@@ -127,8 +127,7 @@ async fn main() -> Result<()> {
             writeln!(
                 bufwriter,
                 "{}",
-                entry
-                    .std_repr(&app.television.channel.prototype.source.output)
+                entry.output(&app.television.channel.prototype.source.output)
             )?;
         }
     }
--- a/television/utils/shell/completion.fish
+++ b/television/utils/shell/completion.fish
@@ -123,7 +123,7 @@ function tv_smart_autocomplete_inline
     # prefix (lhs of cursor)
     set -l current_prompt (commandline --current-process)
 
-    if set -l result (tv $dir --autocomplete-prompt "$current_prompt" --input $tv_query --height 20 3>&1 1>&2 2>&3)
+    if set -l result (tv $dir --autocomplete-prompt "$current_prompt" --input $tv_query --inline)
         # Remove last token from commandline.
         commandline -t ''
 
--- a/television/utils/shell/completion.fish
+++ b/television/utils/shell/completion.fish
@@ -95,34 +95,6 @@ function tv_smart_autocomplete
     # prefix (lhs of cursor)
     set -l current_prompt (commandline --current-process)
 
-    if set -l result (tv $dir --autocomplete-prompt "$current_prompt" --input $tv_query)
-        # Remove last token from commandline.
-        commandline -t ''
-
-        # If dir is the current directory, i.e. './' , clear it.
-        # If the pattern './foo' './bar' instead of 'foo' 'bar' is desired then comment out the check below
-        if test "$dir" = "./"
-            set dir ""
-        end
-
-        for i in $result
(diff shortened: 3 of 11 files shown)


lalvarezt:
waiting for #589

lalvarezt:
@alexpasmantier please have a look, makes the shell_integration usable with a partial height

EDIT: unfortunately we have to make this dance because crossterm Inline is a second class citizen, perhaps it will improve in the future.

television/channels/entry.rs
```diff
@@ -126,7 +126,7 @@ impl Entry {
         self.display.as_deref().unwrap_or(&self.raw)
     }
 
-    pub fn stdout_repr(&self, template: &Option<Template>) -> String {
+    pub fn std_repr(&self, template: &Option<Template>) -> String {
```
alexpasmantier: I'm thinking we might as well change this to `output` since that corresponds to the prototype naming.

television/main.rs
```diff
@@ -120,16 +120,15 @@ async fn main() -> Result<()> {
     debug!("Running application...");
     let output = app.run(stdout().is_terminal(), false).await?;
     info!("App output: {:?}", output);
-    let stdout_handle = stdout().lock();
-    let mut bufwriter = BufWriter::new(stdout_handle);
+    let stderr_handle = stderr().lock();
+    let mut bufwriter = BufWriter::new(stderr_handle);
```
alexpasmantier: Not sure I get this.
We do want to write to stdout by default not stderr 🤔 Otherwise piping tv's output to another command will never work unless explicitly redirecting the streams.

television/utils/shell/completion.fish
```diff
@@ -115,6 +115,34 @@ function tv_smart_autocomplete
...
 
+function tv_smart_autocomplete_inline
+    set -l commandline (__tv_parse_commandline)
+    set -lx dir $commandline[1]
+    set -l tv_query $commandline[2]
+
+    # prefix (lhs of cursor)
+    set -l current_prompt (commandline --current-process)
+
+    if set -l result (tv $dir --autocomplete-prompt "$current_prompt" --input $tv_query --height 20 3>&1 1>&2 2>&3)
```
alexpasmantier: you're redirecting stderr to stdout and stdout to stderr here which shouldn't be necessary if we were using stdout above 🤔

alexpasmantier: Doesn't this swapping break the other shell completion scripts ?

television/main.rs
lalvarezt: not really, this is only for not full-screen apps.

If you do (full-screen version)
```
lalvarezt@FedoraLinux42-wsl:~/git-repos/television$ cargo run -q -- files | echo -n
CHANGELOG.md
```
it works just fine, same as
```
lalvarezt@FedoraLinux42-wsl:~/git-repos/television$ RESULT=$(cargo run -q -- files); echo -n $RESULT
CHANGELOG.md
```

just give it a try

lalvarezt: on the other hand, this doesn't work (but this is not related to the my PR)

```
lalvarezt@FedoraLinux42-wsl:~/git-repos/television$ RESULT=$(cargo run -q -- files --height 20); echo $RESULT
Backtrace (most recent call last):
<...>
The application panicked (crashed).
  Failed to create TUI: The cursor position could not be read within a normal duration
in /home/lalvarezt/git-repos/television/television/render.rs:91
thread: tokio-runtime-worker
```

but this is related to https://github.com/crossterm-rs/crossterm/pull/957, I tried that PR locally and it solves partially the issue

television/utils/shell/completion.fish
lalvarezt: no, if you check the name of the function it is different, we offer two versions, the regular one and this other one if you want to have a partial interface, it's opt-in

tv_smart_autocomplete (original, works for full screen without redirectio)
tv_smart_autocomplete_inline (new one, if you want to have partial screens then the swapping is needed. this is what they do in atuin, got the idea from their implementation)

television/main.rs
alexpasmantier: Added a [simple test](https://github.com/alexpasmantier/television/blob/d8a48fe0ba08e785a37b586f641413a0b51772fa/tests/cli/cli_special.rs#L109-L142) that makes sure piping tv is working correctly.

alexpasmantier: That way you can rebase and make sure this works 👍🏻

television/utils/shell/completion.fish
lalvarezt: > you're redirecting stderr to stdout and stdout to stderr here which shouldn't be necessary if we were using stdout above 🤔

that would be ideal, but I didn't manage to make it work, only the inversion approach renders the interface (stderr) and selects the output (stdout), maybe you have an idea an can make it work but this is very finicky

alexpasmantier: Oh wow I'm tired, my bad. I actually didn't realize we already support fish and thought this was a new feature 😅

alexpasmantier: In that case should we update the bindings below to use that inline version instead?

lalvarezt: for that we should update the zsh and bash versions and test it in other systems, I don't have access to a Mac

alexpasmantier: Ah, I do, will test it 👍🏻

television/main.rs
alexpasmantier: Hmmm the test passes fine locally but won't pass in CI 😢

alexpasmantier: Ended up disabling it in CI altogether since it's really a mess to deal with subprocess handling using portable-pty at the moment

alexpasmantier: > not really, this is only for not full-screen apps.
>
> If you do (full-screen version)
> 
> lalvarezt@FedoraLinux42-wsl:~/git-repos/television$ cargo run -q -- files | echo -n
CHANGELOG.md

Piping into `echo` should fail with an OS error since that's not a thing 😅

lalvarezt: it was a dumb example, I should have used `cat`

I guess wsl has it's own rules 😅, this sometimes works

image

alexpasmantier: I'm unable to get this to work with zsh, whether we're printing to stdout or stderr 🤔 
```
Failed to create TUI: Failed to get cursor position: The cursor position could not be read within a normal duration
```

which seems to be that crossterm issue 😢

alexpasmantier: Actually the only solution I can think of to overcome that crossterm issue would be to capture the cursor position beforehand and pass it via tv's cli but that's getting really nasty...

alexpasmantier: Another solution could be to use another terminal backend for macOS and Linux (e.g. [termion](https://docs.rs/termion/latest/termion/)) and only support `--inline` for macOS and Linux. (and keeping crossterm for windows but disabling the feature)

lalvarezt: I also thought about that possibility, they might not have the same issue

alexpasmantier: I'll branch off of this branch and try a quick migration to termion. Just to see how difficult it would be and whether it solves the problem or not.

alexpasmantier:
@lalvarezt Found a working solution https://github.com/alexpasmantier/television/pull/588/commits/cbf50dde95d262340a8c65eb339915e46aeb0d43

Side note: Also figured that updating [this](https://github.com/alexpasmantier/television/pull/588/files#diff-0a6f826497d8c73ba0dd8d0e06f0be35355417fb5bb7858bc4849a49ca8277b1R69-R71) makes the UX a bit nicer for zsh.

Would you mind giving this a try and telling me if it works as expected on your machine?

lalvarezt:
oh nice 👏, but I'm seeing this weird new line again (in fish), specifically when right at the bottom of the screen

inline

And this in zsh

```
on   ➜ zsh
FedoraLinux42-wsl% source ./television/utils/shell/completion.zsh
FedoraLinux42-wsl% nvim
Error: The cursor position could not be read within a normal duration
FedoraLinux42-wsl% nvim $'\033'\[6n
```

lalvarezt:
made the resizing a much simpler implementation, yes we lose the context from the shell previous commands but being honest while resizing it's not a big deal, IMO
This way it works just as the full screen version and we don't have to manually adjust for every situation

alexpasmantier:
For the cursor position error:

Are you sure you're recompiling the **release** binary ? Because that's what's being used by the shell script.

alexpasmantier:
Regarding resizing:

Was the actual implementation broken? I don't feel we gain that much simplicity by dropping the feature so might as well keep it if it's working as intended.

lalvarezt:
> For the cursor position error:
> 
> Are you sure you're recompiling the **release** binary ? Because that's what's being used by the shell script. 

I patched the zsh file to use the actual implementation, yes

Edit: as you can see I don't use zsh anymore so this just the plain installation with nothing else in it

Edit 2: there's also the thing of the passthrough escape code. For some reason is not being processed

lalvarezt:
> Regarding resizing:
> 
> Was the actual implementation broken? I don't feel we gain that much simplicity by dropping the feature so might as well keep it if it's working as intended.

Yes I'm seeing virtual artifacts. Not sure if that's visible with asciinema. You can try on your side to replicate, open a new terminal session with the screen cleared (the term emulator not in full-screen), open tv --height 20, then horizontally expand. The lines above the prompt get mangled. If you expand vertically (expand) works ok from what I've seen.

Edit: I'm using Wezterm nightly

alexpasmantier:
Screenshot From 2025-07-06 12-01-13

https://github.com/user-attachments/assets/0218190d-c4e8-4870-a648-36d53192d804

I'm getting the extra line too (will take a look at why that's happening), but the `--inline` is working as expected with the **zsh** shell integration.

television/main.rs
lalvarezt: closing this thread since you solved this in https://github.com/alexpasmantier/television/commit/cbf50dde95d262340a8c65eb339915e46aeb0d43 👏

alexpasmantier:
There's also a bug in the zsh script where leaving tv without selecting anything still introduces an empty completion item

lalvarezt:
> There's also a bug in the zsh script where leaving tv without selecting anything still introduces an empty completion item

interesting, for in zsh, I don't get the neither the newlines (not even close to the bottom edge) nor the empty selection

https://github.com/user-attachments/assets/4cd9eff4-880d-4961-81c6-0dbdb8730316

alexpasmantier:
Ah okay, same here, nevermind might have gotten mixed up 👍🏻 

So in the current state of things I believe everything is working as intended right?

The only minor thing left to do being to figure out how to keep the current prompt when using `--inline` without having the newline issue.

lalvarezt:
@alexpasmantier pushed some tweaks, seem to work on my side in all conditions, can you please check

alexpasmantier:
@lalvarezt works perfectly on my side 🎉 

(except for the fact that tv overwrites the prompt which isn't the best UX)

https://github.com/user-attachments/assets/1a4385f6-4f51-43b3-a1a1-4d02e0ff7eb7

lalvarezt:
> @lalvarezt works perfectly on my side 🎉 
> 
> 
> 
> (except for the fact that tv overwrites the prompt which isn't the best UX)
> 
> 
> 
> https://github.com/user-attachments/assets/1a4385f6-4f51-43b3-a1a1-4d02e0ff7eb7
> 
> 

Ah, yes. Since I use a two line prompt I didn't think too much about it. I'll have a look later

alexpasmantier:
Let's merge this then, good job that was a tough one 👍🏻

PR #589, opened 2025-07-03
refactor: more stable behavior for `--inline`, `--width` and `--height`

### LINKED ISSUE #587: `--height` panics on window resize
**Description**
using `tv --height 50` or `tv --inline` panics when resizing the terminal window.

**Example**
```
tv --inline
```
then <kbd>ctrl-+</kbd>

**Expected behavior**
Adapting the height/width automatically depending on available space

**Actual behavior**

https://github.com/user-attachments/assets/fabc1ae1-caa4-4c67-9c46-f5b7c412c216

**Environment**
 - OS: arch linux

alexpasmantier:
Fixes #587

### DIFF
--- a/television/tui.rs
+++ b/television/tui.rs
@@ -1,5 +1,5 @@
 use std::{
-    io::{LineWriter, Write, stderr},
+    io::Write,
     ops::{Deref, DerefMut},
 };
 
@@ -116,15 +116,14 @@ where
 
     pub fn enter(&mut self) -> Result<()> {
         enable_raw_mode()?;
-        let mut buffered_stderr = LineWriter::new(stderr());
+        let backend = self.terminal.backend_mut();
+
+        execute!(backend, EnableMouseCapture)?;
 
         if self.viewport == Viewport::Fullscreen {
-            execute!(buffered_stderr, EnterAlternateScreen)?;
+            execute!(backend, EnterAlternateScreen)?;
             self.terminal.clear()?;
         }
-
-        execute!(buffered_stderr, EnableMouseCapture)?;
-
         Ok(())
     }
 
--- a/television/tui.rs
+++ b/television/tui.rs
@@ -1,5 +1,5 @@
 use std::{
-    io::{LineWriter, StdoutLock, Write, stderr, stdout},
+    io::{LineWriter, Write, stderr},
     ops::{Deref, DerefMut},
 };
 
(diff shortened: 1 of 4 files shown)


alexpasmantier:
@lalvarezt This is a bit cleaner and is more robust when resizing, etc.

Unfortunately, I can't seem to fix the `height_only` test (probably because ratatui tries to call `cursor_position` internally) 🤔

lalvarezt:
let me have a look

lalvarezt:
I cannot commit on this PR, but if we treat (--height) and (--height and --width) using the same logic then it works
[0001-fix-tui-treat-height-and-height-and-width-using-the-.patch](https://github.com/user-attachments/files/21041774/0001-fix-tui-treat-height-and-height-and-width-using-the-.patch)

EDIT: But in that case we lose the Inline logic 🤔, do you find the trade-off to be worth it? Is is more stable to use inline even if we lose for now the possibility of testing it?

television/tui.rs
```diff
@@ -36,94 +35,118 @@ impl<W> Tui<W>
...
+            }
+            (Some(w), Some(h)) => {
+                // get cursor position
+                let mut cursor_pos = Self::cursor_position()?;
+                let term_size = crossterm::terminal::size()?;
+                // scroll if we don't have enough space
+                if cursor_pos.1 + h > term_size.1 {
+                    execute!(
+                        backend,
+                        ScrollUp(cursor_pos.1 + h - term_size.1)
```
lalvarezt, line 61: there's an off-by-one here, replace with
```
ScrollUp(cursor_pos.1 + h - term_size.1 + 1)
```

television/tui.rs
```diff
@@ -36,94 +35,118 @@ impl<W> Tui<W>
...
+                    )?;
+                    cursor_pos.1 = term_size.1.saturating_sub(h);
```
lalvarezt, line 63: same here
```
cursor_pos.1 = term_size.1.saturating_sub(h + 1);
```

television/tui.rs
```diff
@@ -36,94 +35,118 @@ impl<W> Tui<W>
...
+                }
+                options.viewport =
+                    Viewport::Fixed(ratatui::layout::Rect::new(
+                        cursor_pos.0,
```
lalvarezt, line 67: this causes issues if when you render the shell-integration because it starts the tui at the cursor position, is better to use `0`

lalvarezt (commented):
some changes for you to consider

television/tui.rs
alexpasmantier: How do you figure? I'm pretty sure this is correct, unless you want an extra blank line after your TUI.

alexpasmantier:
> EDIT: But in that case we lose the Inline logic 🤔, do you find the trade-off to be worth it? Is is more stable to use inline even if we lose for now the possibility of testing it?

Yeah, maybe we should disable the test for now since `inline` is builtin to ratatui and tested on their side (at least until someone finds a good solution).

television/tui.rs
lalvarezt: uhm, I cannot replicate it anymore. I was getting a blank line after terminal when calling the shell integration but now I don't see it 👹.

television/tui.rs
lalvarezt: ignore this one as well, it it appears again we can check it

lalvarezt:
let's go with this for now, after you merge I'll rebase the other PR and we will have a working shell integration

PR #590, opened 2025-07-06
Fixing bat error

Ktoks:
Adding `'` to bat to prevent errors in preview for Unix cable dotfiles and files configurations.


alexpasmantier:
Thanks for the fix 🙏🏻

PR #605, opened 2025-07-08
docs: Made the landing page more responsive

kapobajza:
@alexpasmantier If you can please check if I've broken anything, regarding the landing page, with this PR.


lalvarezt:
I guess we should add `pnpm` to the prerequisites in the workflow

image

kapobajza:
> I guess we should add `pnpm` to the prerequisites in the workflow
> 
> image

I guess this workflow was just used for testing purposes, the correct one is [deploy-docs.yml](https://github.com/alexpasmantier/television/blob/e5654fcddee638905a5e6478763e6b738ec469d1/.github/workflows/deploy-docs.yml). Correct me if I am wrong, @alexpasmantier

lalvarezt:
> > I guess we should add `pnpm` to the prerequisites in the workflow
> > image
> 
> I guess this workflow was just used for testing purposes, the correct one is [deploy-docs.yml](https://github.com/alexpasmantier/television/blob/e5654fcddee638905a5e6478763e6b738ec469d1/.github/workflows/deploy-docs.yml). Correct me if I am wrong, @alexpasmantier

maybe so but something is failing right now 😅, we could just remove the legacy one

alexpasmantier:
Yup sorry for that, added the pnpm installation step on main. Rebasing this should solve it 😊

alexpasmantier:
Very cool @kapobajza 🙌
Thanks for the PR
Merging this 🎉

Issue #609, opened 2025-07-09
--inline does not behave as expected when there's not enough space at the bottom


lalvarezt:
> I just installed television 0.12.0 and the default ctrl+r bind doesn't work when there isn't enough space at the bottom of my terminal
> 
> after clearing my screen:
> <img width="417" height="397" alt="Image" src="https://github.com/user-attachments/assets/f928cde6-06dd-4f68-a4d3-43bdee20bca9" />
> with no space at the bottom of my terminal:
> <img width="792" height="535" alt="Image" src="https://github.com/user-attachments/assets/af6f5aa1-9155-439f-ad1d-7247154d9409" />
> you can see it tried to make space but it's not using it
> invoking `tv --inline` has this problem, but `tv --height 20` does not
> resizing the terminal while `--inline` is running fixes the issue
> 
> in general, I'll say I'm pretty confused about what `--inline` does. it says
>  If there is insufficient space to meet the minimum height the terminal will scroll.
> 
> but what is the "minimum height"? you can't use `--height` with `--inline`
> 
> ---
> 
> I'm on windows terminal 1.22.111410.0
> ```
> $ zsh --version
> zsh 5.9 (x86_64-debian-linux-gnu)
> $ tv --version
> television 0.12.0
> ``` 

 _Originally posted by @raylu in [#586](https://github.com/alexpasmantier/television/issues/586#issuecomment-3051133288)_

lalvarezt:
I can replicate it, let me have a look

alexpasmantier:
Fixed by #611

raylu:
so speedy! thanks!

should
> If there is insufficient space to meet the minimum height the terminal will scroll.

be updated to say "minimum height (15)" or something?

lalvarezt:
> so speedy! thanks!
> 
> should
> > If there is insufficient space to meet the minimum height the terminal will scroll.
> 
> be updated to say "minimum height (15)" or something?

I also updated the docs. Can you check again?

raylu:
oh sorry, didn't notice the doc update in the PR. thanks!

PR #611, opened 2025-07-09
fix(tui): fix incorrect height calculation when there is not enough space

lalvarezt:
## 📺 PR Description

Solves #609

- The calculation of available_height for the Inline and Fixed mode were updated to correctly account for the remaining space and the necessary height needed for the TUI
- QoL: Refactored the viewport calculation logic for both Inline and Fixed modes to use a shared scrolling function (handle_viewport_scrolling).
- Docs: Added note about default minimum height

## Checklist


alexpasmantier (approved):
Nice, thanks for fixing that 🙏🏻 

Merging this 🎉

PR #619, opened 2025-07-10
fix(cable): quote shell arguments and fix line endings for linux files

### LINKED ISSUE #616: Need quotes around dir placeholders
With latest television 0.12.1 (which looks great, by the way, now with the improved interface), when I use the dirs channel, i get errors in the previewer for directories containing spaces (it splits the dir name using space as delimiters and it complains that the segments do not exist).

The issue is the lack of apostrophes around the {} placeholder.

Changing the preview command to: "ls -la --color=always '{}'" instead of "ls -la --color=always {}" fixes the issue.

I checked the other cables: files and dotfiles contain the apostrophes, but git-repos does not (in two places) and git-diff also does not.

lalvarezt:
## 📺 PR Description

Fixes #616 
Added quotes to prevent shell argument escape

## Checklist


cable/unix/files.toml
```diff
@@ -1,14 +1,14 @@
...
+name = "files"
+description = "A channel to select files and directories"
+requirements = ["fd", "bat"]
+
+[source]
+command = ["fd -t f", "fd -t f -H"]
+
+[preview]
+command = "bat -n --color=always '{}'"
+env = { BAT_THEME = "ansi" }
```
alexpasmantier, line 11: let's remove this too, it was really just there to show the `env` option

alexpasmantier: (in another PR)

alexpasmantier:
Thanks for taking care of that 🙏🏻

PR #620, opened 2025-07-10
fix(tui): account for cursor position when determining the size

### LINKED ISSUE #615: status bar overlaps bottom of screen when non-fullscreen
**Description**
without any flags, `tv` enters fullscreen mode. pressing F12 twice hides and shows the status bar and everything is fine

with `input_bar_position = 'top'` (the default), `tv --height 20` or `tv --inline` draws the status bar over the last line
Image
pressing F12 twice causes it to draw over the box/grid characters
Image

with `input_bar_position = 'bottom'`, the situation is much worse
Image
pressing F12 *once* to hide the status bar collapses the input bar
Image

**Expected behavior**
draw area for other widgets is pushed up when status bar is shown

**Environment**

- OS: win11 + WSL + debian 13
- TV version: 0.12.1
- shell: zsh
- terminal: windows terminal 1.22.11141.0
- **Your configuration file (`~/.config/television/config.toml`)**
```toml
[ui]
use_nerd_font_icons = true
preview_size = 50
input_bar_position = 'bottom'
```
- **The channel you are using** (if applicable and different from the default): files

lalvarezt:
## 📺 PR Description

Fixes #615 

When we want to draw a viewport of `required_height` lines, we need
- required_height -> the lines that belong to the TUI
- 1 -> the cursor’s own line (top border of the viewport)

A few lines later we compute how many rows to scroll:
The extra - 1 deliberately discounts the cursor’s current line (we don’t want to scroll past it), so the earlier + 1 is needed to keep the math balanced.

In short: the + 1 compensates for the later - 1, making sure we always create one full blank line of space for the status bar/prompt and preventing the overlap when the terminal doesn’t have enough free lines left.

## Checklist

### DIFF
--- a/television/tui.rs
+++ b/television/tui.rs
@@ -143,6 +143,9 @@ where
             terminal_size.height, available_height, cursor_position.y
         );
 
+        // We need to add one to the required height to account for the cursor position.
+        let required_height = required_height + 1;
+
         // If we don't have enough space for the required height we need to scroll up.
         if available_height < required_height {
             // Minus one to account for the cursor position.


alexpasmantier (approved):
Thanks 🙏

PR #634, opened 2025-07-12
fix(nix): set meta.mainProgram field

samirettali:
`meta.mainProgram` must be set otherwise `lib.getExe` assumes that the package name is `television` and home manager shell integrations are broken

## Checklist


alexpasmantier:
Thanks for the fix 🙏🏻 

Merging this 🎉

PR #659, opened 2025-07-20
feat: introduction of a new binding system to better handle key/events/actions

lalvarezt:
## 📺 PR Description

the system is backwards compatible with the existing TOML keybinding implementation allowing for slow migration to use the new features

## Checklist

### DIFF
--- a/docs/01-Users/04-keybindings.md
+++ b/docs/01-Users/04-keybindings.md
@@ -1,5 +1,12 @@
 # Keybindings
 
+Television supports two keybinding configuration formats:
+
+1. **Binding syntax** - The modern, recommended format (key => action;)
+2. **TOML format** - The legacy format (action = ["key1", "key2"])
+
+Both formats are fully supported and can be used together. The binding syntax is recommended for new configurations due to its enhanced features like action blocks, event bindings, and channel-specific overrides.
+
 ## Default Keybindings
 
 Default keybindings are as follows:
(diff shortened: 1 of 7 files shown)


lalvarezt:
@alexpasmantier I've added the foundational commits that describe the feature and formalizes the grammar. I'll slowly start committing the implementation as soon as I clean my local copy 😅
wdyt?

docs/01-Users/04-keybindings.md
```diff
@@ -1,41 +1,321 @@
...
+    // Move to next entry and select the next two items
+    f5 => {
+        select_next_entry;
+        toggle_selection_down;
+        toggle_selection_down;
+    };
+}
+```
+
+> **Note**: Action blocks execute each action in order, one after another. This guarantees consistent behavior and prevents actions from interfering with each other.
```
alexpasmantier, line 110: I'm in love with this feature

alexpasmantier:
I love the overall idea of having a more expressive keybinding system and the features you list in the PR are all things I think would be really nice to have in tv.

That being said, I do feel like all the logic behind these features could just as well use TOML as a serialization format which would allow to:
- keep a single configuration file
- keep a single configuration format for everything
- use the same expressive options inside of channel prototypes (which would replace the need for less natural (my opinion only) `for_channel(...)` bindings and the sort.

lalvarezt:
EDIT: added more details in #606

lalvarezt:
closing in favor of #666

Issue #641, opened 2025-07-13
having `--inline` (and `--height`, `width`) as a ui channel option


lalvarezt:
> @lalvarezt I feel like having `--inline` (and `--height`, `width`) as a ui channel option as well would be nice, wdyt? 🤔

 _Originally posted by @alexpasmantier in [#630](https://github.com/alexpasmantier/television/issues/630#issuecomment-3065068022)_

PR #661, opened 2025-07-20
proof of concept: allow passing --inline from channel config

ascandella:
## 📺 PR Description

Partially implements: https://github.com/alexpasmantier/television/issues/641

This is a first draft at implementing UI configuration from channel configs

I haven't implemented tests, nor have I implemented width/height yet. Wanted to get feedback on whether this approach was acceptable first. If you're down with this approach, I'll add tests and implement the rest

Demo:

demo

## Checklist

- [ ] my commits **and PR title** follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format


television/main.rs
```diff
@@ -78,6 +78,9 @@ async fn main() -> Result<()> {
     let channel_prototype =
         determine_channel(&args, &config, readable_stdin, Some(&cable));
 
+    // allow channel to override CLI arguments
+    apply_channel_overrides(&mut args, &channel_prototype);
+
```
alexpasmantier, line 83: I don't think we need this. The cli should always take precedence on the channel prototype.

alexpasmantier:
First of all thanks so much for showing interest in the project. 

Putting the implementation aside, I'm actually really curious to have feedback on the idea itself which I'm not yet entirely convinced of.

Having configurable `--inline` per-channel could mean several things too:
- do we honor the setting each time a user changes channels from the remote? (which implies potentially switching back and forth between terminal modes with the alternate screen)
- or does the setting only work when the channel is invoked via the cli?

This can get rather tricky and I'm not sure it really brings any real user value either when I think of it... 🤔 

wdyt?

ascandella:
Great questions. I'm not too sure how other people use `tv`, but I can speak to my usage and my thoughts on this.

I don't use the channel remote at all-- I use `tv` the same way I use `telescope` inside neovim, where I know what kind of search I'm going to perform before I launch the tool. And depending on the search type, I have a specific set of configuration I use that differs. [Here is my config](https://github.com/ascandella/dotfiles/blob/80971c2d0ac75d58d7fa0617774d90a4f0cb0845/modules/home/files/nvim/config/lua/ai/telescope-config.lua#L99-L148) for telescope.

Conceptually I'd like to use `tv` the same way. Of course I could define shell aliases that launch it with the correct options based on the type of search I'm doing, but it feels like having the ability to say "when I'm searching git branches, I know I'm not going to need fullscreen so don't even try".

As for whether we honor that when switching: I'd vote no. Not only would that be visually jarring but it seems a bit counterintuitive to me, not to mention the extra code it would mean supporting.

But I'm a relatively new `tv` user, so I'm interested in hearing your thoughts (as well as any others that come across this).

alexpasmantier:
I agree with the above. Let's leave this open for now and maybe start a discussion on GitHub or on discord to collect feedback from other users to get it right. Anyways, the implementation looks fine and should be close enough to the target if we decide to proceed with that 👍

alexpasmantier:
Oh and thanks so much for the sponsorship, it means a lot 🙂🙏

PR #662, opened 2025-07-21
fix(website): Fixed broken links on home page

kapobajza:
## 📺 PR Description

Currently the links on the docs homepage are broken, because when we run the website locally it uses [just a slash as the `prefix`/`baseUrl`, but when run as a production build it uses `/television` as the `baseUrl`](https://github.com/alexpasmantier/television/blob/1d6b996c834bf9b2ec7f1811e202c02b4507096c/website/docusaurus.config.ts#L21). So I added a hook that will take the `baseUrl` and append it to the link you provide. This hook should be used for any relative link that we add to the app in React (`.tsx`) files.

## Checklist


alexpasmantier (approved):
Thanks for the fix 🙏🏻

PR #666, opened 2025-07-22
Binding foundation

lalvarezt:
## 📺 PR Description

**IMPORTANT** Due to the nature of the changes this is a backwards incompatible change

This PR reworks the already existing Action -> Bindings system in to a more scalable system using
- Keys -> Actions
- Events -> Actions

With the new system a key or event can execute one or more actions sequentially, for example

```toml
[keybindings]
# Zen Mode
"f1" = ["toggle_preview", "toggle_status_bar"]
```

Exposes input edit actions that we previously hard-coded

## Checklist

### DIFF
--- a/television/action.rs
+++ b/television/action.rs
@@ -1,4 +1,5 @@
 use serde::{Deserialize, Serialize};
+use serde_with::{OneOrMany, serde_as};
 use std::fmt::Display;
 
 /// The different actions that can be performed by the application.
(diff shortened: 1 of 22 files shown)


television/action.rs
alexpasmantier, line 1: I'm pretty sure all this can be avoided by using serde_with's `OneOrMany` don't you think?

television/config/keybindings.rs
```diff
@@ -10,30 +10,32 @@ use std::hash::Hash;
...
-                    .map(std::string::ToString::to_string)
-                    .collect();
-                write!(f, "{}", keys_str.join(", "))
-            }
-        }
-    }
-}
 
 /// Generic bindings structure that can map any key type to actions
+/// Generic bindings structure that maps any key type to actions.
```
alexpasmantier: I think you have an extra line here

television/config/keybindings.rs
```diff
@@ -53,23 +55,104 @@ where
...
+/// use std::str::FromStr;
+///
+/// let event = EventType::from_str("mouse-click").unwrap();
+/// assert_eq!(event, EventType::MouseClick);
+///
+/// let custom = EventType::Custom("my-event".to_string());
+/// assert_eq!(custom.to_string(), "my-event");
+/// ```
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize)]
 #[serde(rename_all = "kebab-case")]
```
alexpasmantier, line 154: we should probably use snake case here to be consistent with actions

lalvarezt: uhm, but keys use kebab-case. I put key/events as the same thing (left side) and actions (snake_case) on the right side. Would you agree?

alexpasmantier: Ah my bad. I tend to think we should use the same rule for every serialized item, but I'm open to alternatives if there's a good reason for them.

television/action.rs
lalvarezt: my understanding is that OneOrMany always returns vec<T> and since most keys have only a single item it would be a waste with a potential performance impact (maybe not that great). Another advantage is that it might be useful later or to separate a single action from a sequence.
any thoughts on this?

television/config/keybindings.rs
alexpasmantier, line 1: could we do something like:

```rs
use serde::{Deserialize, Deserializer};

#[derive(Debug)]
pub enum EventType {
    MouseClick,
    MouseScrollUp,
    MouseScrollDown,
    Resize,
    Custom(String),
}

impl<'de> Deserialize<'de> for EventType {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let s = String::deserialize(deserializer)?;
        let event = match s.as_str() {
            "mouse-click" => EventType::MouseClick,
            "mouse-scroll-up" => EventType::MouseScrollUp,
            "mouse-scroll-down" => EventType::MouseScrollDown,
            "resize" => EventType::Resize,
            other => EventType::Custom(other.to_string()),
        };
        Ok(event)
    }
}
```
to avoid having to implement `FromStr` manually here?


PS: for some reason github won't let me comment the corresponding lines.

television/action.rs
alexpasmantier: rust always keeps enough space available for an enum's largest variant, so 
```rs
#[derive(
    Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash, PartialOrd, Ord,
)]
#[serde(untagged)]
pub enum Actions {
    /// A single action binding
    Single(Action),
    /// Multiple actions executed in sequence
    Multiple(Vec<Action>),
}
```
will always end up taking enough space to hold `Multiple(Vec<Action>)`.

Now the only real difference here would be if `Action` didn't hold any heap-allocated variants, in which case using the intermediate struct would save us from allocating to the heap. But since that's not the case we still need to allocate to the heap and the performance tradeoff here is negligible.

lalvarezt: you have a point there, pushed a commit with the changes

television/config/keybindings.rs
lalvarezt: having issues with this one, the refactor gets complicated and the compiler starts bitching, can you have a look. Is the tradeoff worth it?

alexpasmantier: just had another look at the code and it's fine tbh 👍🏻

alexpasmantier:
@lalvarezt just read through the code and I think I'm pretty happy with the PR as it is now. Any things you wish to add/update before merging?

lalvarezt:
No, I think it's a good start, events are basically a placeholder  but it can be improved later on

PR #668, opened 2025-07-24
Rework help panel

lalvarezt:
## 📺 PR Description

Rework the help panel to show live keybinding source information (global, channel) instead of hard-coded entries

Before
image

After
image

### Command

```bash
RUST_LOG=debug cargo run -q -- --cable-dir ./cable/unix --config-file ./.config/config.toml --show-help-panel --keybindings "f10 = \"toggle_help\"; f9 = \"toggle_status_bar\""
```

### Config file
```toml
[metadata]
name = "files"
description = "A channel to select files and directories"
requirements = ["fd", "bat"]

[source]

[preview]
command = "bat -n --color=always '{}'"
env = { BAT_THEME = "ansi" }

[keybindings]
shortcut = "f1"
f9 = "toggle_preview"
f8 = "quit"
esc = false
```

## Checklist


television/config/mod.rs
```diff
@@ -236,15 +250,200 @@ impl Config {
...
+
+        debug!(
+            "Applying {} CLI keybinding overrides",
+            cli_keybindings.len()
+        );
+
+        let mut override_count = 0;
+        let mut new_key_count = 0;
+        let mut preserved_global = 0;
+        let mut preserved_channel = 0;
```
alexpasmantier: I feel like a more helpful debug message could be:
```
keybindings before: {...}
keybindings after: {...}
```
This avoids the extra count logic which is really only useful for debug and trace.

television/config/keybindings.rs
```diff
@@ -279,21 +279,119 @@ where
...
+///
+/// # Variants
+///
+/// * `Global` - Bindings from user configuration files or default settings
+/// * `Channel` - Bindings specific to individual channels/data sources
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BindingSource {
+    Global,
+    Channel,
+}
```
alexpasmantier: Let's use `BindingOrigin` here or smthg like that to avoid the semantic confusion with "source".

television/config/keybindings.rs
```diff
@@ -307,23 +405,88 @@ where
...
+        if let Ok(tv_key) = Key::from_str(&key.to_string()) {
+            keybinding_source.add_global_key(tv_key);
+            trace!("Marked existing key '{}' as global", key);
+        }
+    }
+
+    // Merge new bindings and mark them based on source type
+    let mut merged_count = 0;
+    let mut override_count = 0;
+
```
alexpasmantier: same remark as earlier here: does the count really bring information that's valuable enough to justify the extra logic?

television/config/keybindings.rs
lalvarezt: BindingType? Scope?

television/config/keybindings.rs
lalvarezt: yeah, I added a lot of intermediate values to be able to debug the thing, let me clean it up a bit

alexpasmantier:
To be honest I'm not too fond of this new extra `KeybindingSource` struct that's getting carried around the code.

A solution might be to make the current `Keybindings` struct carry richer information 🤔 but still expose its current convenience methods (e.g. getting an action based on a key without worrying about where the mapping came from).

Or maybe not 😁 but I can't dive in right now so I'll leave it up to you.

television/config/keybindings.rs
alexpasmantier: tough one 😅

lalvarezt:
> To be honest I'm not too fond of this new extra `KeybindingSource` struct that's getting carried around the code.
> 
> A solution might be to make the current `Keybindings` struct carry richer information 🤔 but still expose its current convenience methods (e.g. getting an action based on a key without worrying about where the mapping came from).
> 
> Or maybe not 😁 but I can't dive in right now so I'll leave it up to you.

I understand, let me play with it a bit

lalvarezt:
> To be honest I'm not too fond of this new extra `KeybindingSource` struct that's getting carried around the code.
> 
> A solution might be to make the current `Keybindings` struct carry richer information 🤔 but still expose its current convenience methods (e.g. getting an action based on a key without worrying about where the mapping came from).
> 
> Or maybe not 😁 but I can't dive in right now so I'll leave it up to you.

I dropped the whole thing entirely, it really didn't add much and made things more complex for no reason

alexpasmantier:
This does look much cleaner! 🙏🏻 Thanks for the refactor.

PR #669, opened 2025-07-27
feat: support for external actions

lalvarezt:
## 📺 PR Description

This PR adds support for definition of external actions, allowing the following

```toml
[metadata]
name = "files"
description = "A channel to select files and directories"
requirements = ["fd", "bat"]

[source]

[preview]
command = "bat -n --color=always '{}'"
env = { BAT_THEME = "ansi" }

[keybindings]
shortcut = "f1"
f8 = "thebatman"
f9 = "lsman"

[actions.thebatman]
description = "cats the file"
command = "bat '{}'"
env = { BAT_THEME = "ansi" }

[actions.lsman]
description = "show stats"
command = "ls '{}'"
```

## Checklist

### DIFF
--- a/cable/unix/files.toml
+++ b/cable/unix/files.toml
@@ -12,3 +12,8 @@ env = { BAT_THEME = "ansi" }
 
 [keybindings]
 shortcut = "f1"
+ctrl-f12 = "edit"
+
+[actions.edit]
+description = "Opens the selected entry with Neovim"
+command = "nvim '{}'"
--- a/television/action.rs
+++ b/television/action.rs
@@ -5,9 +5,7 @@ use std::fmt::Display;
 use crate::event::Key;
 
 /// The different actions that can be performed by the application.
-#[derive(
-    Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash, PartialOrd, Ord,
-)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Hash, PartialOrd, Ord)]
 #[serde(rename_all = "snake_case")]
 pub enum Action {
     // input actions
@@ -118,6 +116,8 @@ pub enum Action {
     /// Handle mouse click event at specific coordinates
     #[serde(skip)]
     MouseClickAt(u16, u16),
+    /// Execute an external action
+    ExternalAction(String),
 }
 
 /// Container for one or more actions that can be executed together.
@@ -365,7 +365,68 @@ impl Display for Action {
                 write!(f, "select_entry_at_position")
             }
             Action::MouseClickAt(_, _) => write!(f, "mouse_click_at"),
+            Action::ExternalAction(name) => write!(f, "{}", name),
         }
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for Action {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
+
+        let action = match s.as_str() {
+            "add_input_char" => Action::AddInputChar(' '),
+            "delete_prev_char" => Action::DeletePrevChar,
+            "delete_prev_word" => Action::DeletePrevWord,
+            "delete_next_char" => Action::DeleteNextChar,
+            "delete_line" => Action::DeleteLine,
+            "go_to_prev_char" => Action::GoToPrevChar,
+            "go_to_next_char" => Action::GoToNextChar,
+            "go_to_input_start" => Action::GoToInputStart,
+            "go_to_input_end" => Action::GoToInputEnd,
+            "render" => Action::Render,
+            "resize" => Action::Resize(0, 0),
+            "clear_screen" => Action::ClearScreen,
+            "toggle_selection_down" => Action::ToggleSelectionDown,
+            "toggle_selection_up" => Action::ToggleSelectionUp,
+            "confirm_selection" => Action::ConfirmSelection,
+            "select_and_exit" => Action::SelectAndExit,
+            "expect" => Action::Expect(Key::Char(' ')),
+            "select_next_entry" => Action::SelectNextEntry,
+            "select_prev_entry" => Action::SelectPrevEntry,
+            "select_next_page" => Action::SelectNextPage,
+            "select_prev_page" => Action::SelectPrevPage,
+            "copy_entry_to_clipboard" => Action::CopyEntryToClipboard,
+            "scroll_preview_up" => Action::ScrollPreviewUp,
+            "scroll_preview_down" => Action::ScrollPreviewDown,
+            "scroll_preview_half_page_up" => Action::ScrollPreviewHalfPageUp,
+            "scroll_preview_half_page_down" => {
+                Action::ScrollPreviewHalfPageDown
+            }
+            "open_entry" => Action::OpenEntry,
+            "tick" => Action::Tick,
+            "suspend" => Action::Suspend,
+            "resume" => Action::Resume,
+            "quit" => Action::Quit,
+            "toggle_remote_control" => Action::ToggleRemoteControl,
+            "toggle_help" => Action::ToggleHelp,
+            "toggle_status_bar" => Action::ToggleStatusBar,
+            "toggle_preview" => Action::TogglePreview,
+            "error" => Action::Error(String::new()),
+            "no_op" => Action::NoOp,
+            "cycle_sources" => Action::CycleSources,
+            "reload_source" => Action::ReloadSource,
+            "switch_to_channel" => Action::SwitchToChannel(String::new()),
+            "watch_timer" => Action::WatchTimer,
+            "select_prev_history" => Action::SelectPrevHistory,
+            "select_next_history" => Action::SelectNextHistory,
+            _ => Action::ExternalAction(s),
+        };
+
+        Ok(action)
     }
 }
 
@@ -460,6 +521,9 @@ impl Action {
             // Mouse actions
             Action::SelectEntryAtPosition(_, _) => "Select at position",
             Action::MouseClickAt(_, _) => "Mouse click",
+
+            // External actions
+            Action::ExternalAction(_) => "External action",
         }
     }
 }
--- a/television/app.rs
+++ b/television/app.rs
@@ -1,7 +1,10 @@
 use crate::{
     action::Action,
     cable::Cable,
-    channels::{entry::Entry, prototypes::ChannelPrototype},
+    channels::{
+        entry::Entry,
+        prototypes::{ActionSpec, ChannelPrototype},
+    },
     cli::PostProcessedCli,
     config::{Config, DEFAULT_PREVIEW_SIZE, default_tick_rate},
     event::{Event, EventLoop, InputEvent, Key, MouseInputEvent},
@@ -147,6 +150,7 @@ pub enum ActionOutcome {
     EntriesWithExpect(FxHashSet<Entry>, Key),
     Input(String),
     None,
+    ExternalAction(ActionSpec, String),
 }
 
 /// The result of the application.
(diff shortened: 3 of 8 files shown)


alexpasmantier:
This is awesome. It's pretty much what I was thinking of too which is great 😁 Here are the other things I had in mind for this summed up in part of a channel prototype:

```toml
# open the selected entry/ies in vim without leaving tv (i.e. exiting vim returns to tv)
# this is calling `fork` behind the scenes
[actions.open-in-vim]
description = "open the selected entry with vim"
command = "vim '{}'"
become = false

# this makes tv become vim (i.e. the tv process turns into vim so leaving vim will
# return you to the shell)
# this is calling `execve` behind the scenes
[actions.become-vim]
description = "exit tv and open the selected entry in vim"
command = "vim '{}'"
become = true

# when you don't care about the command's output
[actions.rm]
description = "delete a file"
command = "rm '{}'"
silent = true  # or 'verbose' or something like that

# when you actually care about the command's output
[actions.ls]
description = "list the entries in the currently selected directory"
command = "ls '{}'"
silent = false
```

These extra features **shouldn't be added directly in this PR**, but we should try to shape the code in a way that makes it easier to add them later on.

I also was thinking of adding a channel-like selection UI for the actions of a given channel, e.g.:
- you're on the files channel and have your matching item
- you want to perform an action that isn't tied to a keybinding and open a popup fuzzy finder with all actions for this channel as well as their description and any other useful metadata
- you start typing, find your action and press <kbd>enter</kbd> to trigger it

The UI would be very similar to what the remote control already looks like. And the code would be very similar too (the remote control is more or less implemented as a hard-coded channel to reuse as much existing logic as possible).

What do you think? 

PS: was considering either using libc directly or https://crates.io/crates/nix (at least for unix) unless there's a simpler way to do the equivalent of `execve` using the standard lib that I'm not aware of 🤔

cable/unix/files.toml
```diff
@@ -12,3 +12,8 @@ env = { BAT_THEME = "ansi" }
 
 [keybindings]
 shortcut = "f1"
+ctrl-f12 = "edit"
```
alexpasmantier: Maybe we should use a prefix here to avoid clashes with tv's builtin actions, e.g.:

```toml
[keybindings]
ctrl-f12 = "actions:edit"
ctrl-f11 = "actions:copy-entry-to-clipboard"  # the action defined below
ctrl-f10 = "copy-entry-to-clipboard"          # tv's builtin Action::CopyEntryToClipboard

[actions.edit]

[actions.copy-entry-to-clipboard]
```

This would also probably re-simplify deserialisation below.

lalvarezt: yup, makes sense, thanks

television/app.rs
```diff
@@ -669,6 +683,58 @@ impl App {
...
+                                action_name
+                            )))?;
+                        } else {
+                            error!("No entry selected for external action");
+                            self.action_tx.send(Action::Error(
+                                "No entry selected for external action"
+                                    .to_string(),
+                            ))?;
+                        }
+                    }
```
alexpasmantier: A couple of things here:
- we possibly need to deal with multiple entries (the user might have selected 3 entries and wishes to execute `rm` on all three)
- we won't always want to exit tv after performing an action (see PR main comment)

television/main.rs
```diff
@@ -111,6 +112,41 @@ async fn main() -> Result<()> {
...
     let output = app.run(stdout().is_terminal(), false).await?;
     info!("App output: {:?}", output);
+
+    // Handle external action execution after terminal cleanup
+    if let Some((action_spec, entry_value)) = output.external_action {
+        debug!("Executing external action command after terminal cleanup");
+
+        // Format the command template with the entry value
+        let formatted_command =
+            action_spec.command.get_nth(0).format(&entry_value)?;
```
alexpasmantier: it kind of feels weird having to call `get_nth(0)` here but I guess that's how we ended up implementing multiple sources per-channel 🤷🏻

not a big deal

television/main.rs
```diff
@@ -111,6 +112,41 @@ async fn main() -> Result<()> {
...
+        // Execute the command with inherited stdio
+        let mut child = shell_command(
+            &formatted_command,
+            action_spec.command.interactive,
+            &action_spec.command.env,
+        )
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .spawn()?;
```
alexpasmantier: This is where we don't always want to:
- inherit std file descriptors
- spawn a new process

lalvarezt:
@alexpasmantier added the changes you requested, let placeholders for future enhancement

television/action.rs
```diff
@@ -365,10 +365,125 @@ impl Display for Action {
...
+            "error" => Action::Error(String::new()),
+            "no_op" => Action::NoOp,
+            "cycle_sources" => Action::CycleSources,
+            "reload_source" => Action::ReloadSource,
+            "switch_to_channel" => Action::SwitchToChannel(String::new()),
+            "watch_timer" => Action::WatchTimer,
+            "select_prev_history" => Action::SelectPrevHistory,
+            "select_next_history" => Action::SelectNextHistory,
+            s if s.starts_with("actions:") => {
+                let action_name = &s[8..]; // Remove "actions:" prefix
```
alexpasmantier: Let's add a constant here:
```rust
const ACTION_PREFIX: &str = "actions";
```
And then trim it from the string

television/channels/prototypes.rs
```diff
@@ -159,6 +159,48 @@ impl CommandSpec {
...
+#[derive(
+    Debug, Clone, Default, serde::Deserialize, serde::Serialize, PartialEq,
+)]
+#[serde(rename_all = "lowercase")]
+pub enum ExecutionMode {
+    /// Fork the command as a child process (current behavior, tv stays open)
+    #[default]
+    Fork,
+    /// Replace the current process with the command (tv exits, command takes over)
+    Become,
```
alexpasmantier: This should probably be `Execute` to reflect the system call

television/channels/prototypes.rs
```diff
@@ -159,6 +159,48 @@ impl CommandSpec {
...
+
+#[derive(Debug, Clone, serde::Deserialize, serde::Serialize, PartialEq)]
+pub struct ActionSpec {
+    #[serde(default)]
+    pub description: Option<String>,
+    #[serde(flatten)]
+    pub command: CommandSpec,
+    /// How to execute the command (fork vs become)
+    #[serde(default)]
+    pub r#become: bool,
```
alexpasmantier: "become" seems cumbersome being a reserved identifier. Let's use something like:
```rs
pub mode: ExecutionMode,
```
(with `ExecutionMode::Fork` being the default)

television/channels/prototypes.rs
```diff
@@ -159,6 +159,48 @@ impl CommandSpec {
...
+
+/// Output handling mode for external actions
+#[derive(
+    Debug, Clone, Default, serde::Deserialize, serde::Serialize, PartialEq,
+)]
+#[serde(rename_all = "lowercase")]
+pub enum OutputMode {
+    /// Inherit stdin/stdout/stderr (current behavior)
+    #[default]
+    Inherit,
```
alexpasmantier: I don't think we need this. The way I see it:
- if `mode` is `Fork`:
  - we either want `Capture` or `Discard`
- if `mode` is `Execute`, the new process takes over and will keep the same file descriptors anyway

television/channels/prototypes.rs
```diff
@@ -159,6 +159,48 @@ impl CommandSpec {
...
+    pub description: Option<String>,
+    #[serde(flatten)]
+    pub command: CommandSpec,
+    /// How to execute the command (fork vs become)
+    #[serde(default)]
+    pub r#become: bool,
+    /// How to handle command output
+    #[serde(default)]
+    pub output_mode: OutputMode,
+}
```
alexpasmantier: could you just add this for later so we don't forget?
```rs
// TODO: add `requirements` (see `prototypes::BinaryRequirement`)
```

television/app.rs
```diff
@@ -669,6 +683,81 @@ impl App {
...
+                            // Use single entry under cursor
+                            std::iter::once(current_entry).collect()
+                        } else {
+                            debug!("No entries available for external action");
+                            self.action_tx.send(Action::Error(
+                                "No entry available for external action"
+                                    .to_string(),
+                            ))?;
+                            return Ok(ActionOutcome::None);
+                        };
```
alexpasmantier: There's an existing method for what you're trying to do here: `television::Television::get_selected_entries`

```rs
let selected_entries = self.television.get_selected_entries();
if selected_entries.is_none() {...}
```

alexpasmantier: or yet simpler:
```rs
if let Some(selected_entries) = self.television.get_selected_entries() {
    ...
}
debug!("No entries available for external action");
...
return Ok(ActionOutcome::None);

**Comment by lalvarezt** [comment]
> PS: was considering either using libc directly or https://crates.io/crates/nix (at least for unix) unless there's a simpler way to do the equivalent of `execve` using the standard lib that I'm not aware of 🤔

Yazi seems to use `tokio:process::Command`
https://github.com/sxyazi/yazi/blob/main/yazi-scheduler/src/process/process.rs

**Comment by alexpasmantier** [comment]
> > PS: was considering either using libc directly or https://crates.io/crates/nix (at least for unix) unless there's a simpler way to do the equivalent of `execve` using the standard lib that I'm not aware of 🤔
> 
> Yazi seems to use `tokio:process::Command` https://github.com/sxyazi/yazi/blob/main/yazi-scheduler/src/process/process.rs

Having a quick look it seems to me like yazi never allows the user to turn the app into another process and instead always uses fork and task management to handle external commands.

**Comment by lalvarezt** [comment]
@alexpasmantier is there anything else you think we should put in this PR, considering that you wanted to leave some content for the future?

**Comment by alexpasmantier** [comment]
The PR looks fine, thanks for the refactors 👍🏻 

I'm doing quite a bit of cleaning/refactoring, mainly on the config because I feel like we've reached a point where things are getting hard to wrap one's head around and I want to lay that down before we add any more features and code, hence me delaying recent PRs.

**Comment by lalvarezt** [comment]
> The PR looks fine, thanks for the refactors 👍🏻 
> 
> 
> 
> I'm doing quite a bit of cleaning/refactoring, mainly on the config because I feel like we've reached a point where things are getting hard to wrap one's head around and I want to lay that down before we add any more features and code, hence me delaying recent PRs.

Yup, I feel you. So many moving parts 😅

**Comment by alexpasmantier** [comment]
Okay, change of plans, the config refactor actually needs to be larger and deeper than I imagined. I'll postpone that for now and work on it on the side.

In the meantime, let's proceed with this.

File: television/app.rs [hunk 11]
```diff
@@ -669,6 +683,51 @@ impl App {
...
+                                    selected_entries
+                                        .iter()
+                                        .map(|entry| {
+                                            let raw = entry.raw.clone();
+                                            if raw
+                                                .chars()
+                                                .any(char::is_whitespace)
+                                            {
+                                                format!("'{}'", raw)
+                                            } else {
```
alexpasmantier: Don't you think we should just by default quote everything regardless of if there's a space here?

lalvarezt: I mean we could, I don't have anything against it. It might be safer that way

alexpasmantier: Also what do you think of moving this logic into a separate function and calling it in `execute_action` instead.

```rs
// where E is an iterator of Entry
fn make_entry_args(entries: E) -> Option<String> {
    // if E empty -> None
    // else quote and concatenate with a whitespace
}

fn execute_action(action: &ActionSpec, entries: &Vec<Entry>) -> Result<ExitStatus> {
    // call make_entry_args in here
}
```

lalvarezt: oh, that's very nice

lalvarezt: made some changes to give the possibility of using Template to tweak the command. Based on this use case I might add a new function `quote` to string_pipeline

alexpasmantier:
That's great! Thanks for the PR !! 🙏

PR #671, opened 2025-07-28
feat: add frecency support

lalvarezt:
## 📺 PR Description

This PR add frecency support to already selected entries (opt-in), they will be boosted over the nucleo matcher results

## Checklist

- [x] my commits **and PR title** follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format
- [ ] if this is a new feature, I have added tests to consolidate the feature and prevent regressions
- [ ] if this is a bug fix, I have added a test that reproduces the bug (if applicable)
- [x] I have added a reasonable amount of documentation to the code where appropriate

PR #672, opened 2025-07-29
feat(channel): disable sorting

raylu:
## 📺 PR Description

allow disabling of sorting, which makes ctrl+r way more useful

this partially addresses #194

## Checklist

### DIFF
--- a/television/app.rs
+++ b/television/app.rs
@@ -18,7 +18,7 @@ use tracing::{debug, error, trace};
 
 #[allow(clippy::struct_excessive_bools)]
 pub struct AppOptions {
-    /// Whether the application should use subsring matching instead of fuzzy matching.
+    /// Whether the application should use substring matching instead of fuzzy matching.
     pub exact: bool,
     /// Whether the application should automatically select the first entry if there is only one
     /// entry available.


television/app.rs
```diff
@@ -18,8 +18,7 @@ use tracing::{debug, error, trace};
 
 #[allow(clippy::struct_excessive_bools)]
 pub struct AppOptions {
-    /// Whether the application should use subsring matching instead of fuzzy
-    /// matching.
+    /// Whether the application should use subsring matching instead of fuzzy matching.
```
lalvarezt: since we're changing this line anyways, can you please fix the typo `subsring`

alexpasmantier:
This looks great, nice job 👍🏻 

Here are my thoughts:

I'm wondering if changing the semantics a bit could be less confusing to users. Although this really **disables** sorting in nucleo, I feel like we might want to hide that from the user and expose flags that reflect the user's intention rather than what's happening behind the scenes.

What I mean by the above is that disabling nucleo's internal sorting is almost always something the user will want to do **when they actually care about sorting** and so passing `--sort=false` in these scenarios feels kind of counter-intuitive.

I would change the `sort=false` flag and prepare for other sorting methods by using something like:
```
--sort='strategy'
```

```rs
enum SortingStrategy {
    #[default]
    Unsorted,  // the current behavior
    Natural,  // disabling nucleo's sorting -> entries keep their natural order 
              // (the order they were streamed in to tv)
    Frecency, 
    Template,  // this could for example extract a score from the entry using a template
}
```

This would also avoid having to do things like:
```rs
    #[serde(default = "default_true")]
    pub prefer_prefix: bool,
    #[serde(default = "default_true")]
    pub sort_results: bool,
}

fn default_true() -> bool {
    true
}
```

alexpasmantier:
See https://github.com/alexpasmantier/television/pull/671#issuecomment-3136361740 for more context.

gauravkumar37:
@alexpasmantier hey, given that config/channel refactor has landed :tada: and this PR is supposed to precede https://github.com/alexpasmantier/television/pull/671 is this the right time to look at this one?
Thanks again for the excellent project, the only thing breaking for me at the moment is non-chronological sort order for shell history channel

man-brain:
Any success here? Looks like it should fix the case for ctrl+r search results sorting.

raylu:
I tried to implement `SortingStrategy` and got stuck
no objection if someone wants to take over

alexpasmantier:
I'm planning to work on this in the coming days. 

The reason for the delay is that I want to do this as a part of a bigger refactoring which might involve a bit more than a one evening session and have been a bit busy the last couple of weeks.

alexpasmantier:
Disabling sorting and frecency sorting are coming in the next release.

Closing this. Thanks for all the investigation work 🙏🏻

PR #675, opened 2025-07-31
feat: better handling external action's arguments

lalvarezt:
## 📺 PR Description

Add the possibility to change the separator when multiple items are selected and we're executing and external action.

```toml
[actions.edit]
description = "Open selected files in editor"
command = "nvim {split:\\n:..|map:{append:'|prepend:'}|join: }"
mode = "execute"
separator = "\n"
# example: inputs "file1" and "file 2" will generate the command
# nvim 'file1' 'file 2'
```

Default value is the newline, but depending on the actual entries it might be beneficial to use another one

Add external actions documentation

Add support for common use case where we only want to escape every selection, allowing a short syntax like:

```toml
command = 'nvim {}'
# for inputs "file1.txt" and "file2.txt" produces
# nvim 'file1.txt' 'file2.txt
# for inputs "file1's.txt" and "file2.txt" produces
# nvim 'file1\'s.txt' 'file2.txt
```

## Checklist

### DIFF
--- a/television/utils/command.rs
+++ b/television/utils/command.rs
@@ -6,12 +6,15 @@ use crate::{
     utils::shell::Shell,
 };
 use anyhow::Result;
+use lazy_regex::{Lazy, Regex, regex};
 use rustc_hash::FxHashSet;
 use std::{
     collections::HashMap,
     process::{Command, ExitStatus, Stdio},
 };
 use tracing::debug;
+
+static COMPLEX_BRACES_REGEX: &Lazy<Regex> = regex!(r"\{[^}]+\}");
 
 #[cfg(not(unix))]
 use tracing::warn;
@@ -96,20 +99,43 @@ pub fn format_command(
         entries.len()
     );
 
-    // Concatenate entries with separator for template processing
-    let entries_str = entries
-        .iter()
-        .map(|entry| entry.raw.as_str())
-        .collect::<Vec<_>>()
-        .join(separator);
-    debug!("Concatenated entries input: {:?}", entries_str);
+    let template_str = template.raw();
 
-    // Process through template system
-    debug!("Using template: {}", template);
-    let formatted_command = template.format(&entries_str)?;
-    debug!("Final command: {:?}", formatted_command);
+    // Check if template has only simple braces (syntactic sugar)
+    let has_only_simple_braces = !COMPLEX_BRACES_REGEX.is_match(template_str);
+    if has_only_simple_braces {
+        // Handle simple braces with predictable multi-value logic
+        debug!(
+            "Using simple brace syntactic sugar for template: {}",
+            template_str
+        );
 
-    Ok(formatted_command)
+        // Multiple entries: quote each and join with spaces
+        let quoted_entries: Vec<String> = entries
+            .iter()
+            .map(|entry| format!("'{}'", entry.raw.replace('\'', r"\'")))
+            .collect();
+        let entries_joined = quoted_entries.join(" ");
+        let formatted_command = template_str.replace("{}", &entries_joined);
+        debug!("Multiple entries command: {:?}", formatted_command);
+        Ok(formatted_command)
+    } else {
+        // Complex braces: use existing template processing
+        debug!("Using complex template processing for: {}", template_str);
+
+        // Concatenate entries with separator for template processing
+        let entries_str = entries
+            .iter()
+            .map(|entry| entry.raw.as_str())
+            .collect::<Vec<_>>()
+            .join(separator);
+        debug!("Concatenated entries input: {:?}", entries_str);
+
+        // Process through template system
+        let formatted_command = template.format(&entries_str)?;
+        debug!("Final command: {:?}", formatted_command);
+        Ok(formatted_command)
+    }
 }
 
 /// Execute an external action with the appropriate execution mode and output handling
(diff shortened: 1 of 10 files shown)


alexpasmantier:
I'm actually wondering if using a `Template` to create multiple entries arguments (joining with space etc.) is really the good design choice.

I think the original rust implementation was the way to go, systematically using the unix convention `cmd 'arg1' 'arg2' 'arg3'` which will almost always be what the user expects to be happening.

Let's not over-complicate this and give it a little time in the next release to get user feedback and ideas instead of adding too many features at once.

lalvarezt:
I understand the simplicity point of view. Just playing devil's advocate here, I guess it all that depends on what you expect to receive on the other side. You might want to:
- not have quotes anywhere
- take a section of the value shown on the results windows
- maybe format as json, csv, etc
- maybe you want to call an operation on something that's inside the selected item, or related to it

Maybe's too niche, just thinking on possible use cases

alexpasmantier:
I'm pretty sure we can do most of the above already using each entry's `output` template, and only having to then use the existing `format_action` to format these correctly to be passed as arguments to a shell command.

lalvarezt:
true, just in that case then the output formatting is linked to all the actions.
In any case, it's your call ✌️

alexpasmantier:
> true, just in that case then the output formatting is linked to all the actions. In any case, it's your call ✌️

That is true...

The main reason behind all this is that I don't want users to have to write `nvim {split:\\n:..|map:{append:'|prepend:'}|join: }` when that's really the default behavior you expect 99% of the time.

Opening nvim for the selection should be as simple as `nvim '{}'` with the single/multiple entries logic being handled for you automaticallly.

lalvarezt:
> > true, just in that case then the output formatting is linked to all the actions. In any case, it's your call ✌️
> 
> That is true...
> 
> The main reason behind all this is that I don't want users to have to write `nvim {split:\\n:..|map:{append:'|prepend:'}|join: }` when that's really the default behavior you expect 99% of the time.
> 
> Opening nvim for the selection should be as simple as `nvim '{}'` with the single/multiple entries logic being handled for you automaticallly.

this is very true, I also find it very verbose 😅. 🤔 What if we treat empty braces as a syntactic sugar that will escape multi-values as expected, and the rest we do like now?

alexpasmantier:
That does sound interesting but could you give an example of what you mean by that? 😄

alexpasmantier:
do you mean syntactic sugar inside string-pipeline or an intermediate conversion inside tv ?

alexpasmantier:
An intermediate conversion inside tv probably makes more sense 🤔

lalvarezt:
> An intermediate conversion inside tv probably makes more sense 🤔

yes, this is what I meant, before entering the string_pipeline (in your case the Template) just make a quick check, if all braces are simple ones, just go with easy predictable, if one of them has something then process it

lalvarezt:
@alexpasmantier what about now

alexpasmantier:
Do you mind if I try something directly onto this PR? 😇

television/utils/command.rs
```diff
@@ -88,26 +92,50 @@ pub fn shell_command<S>(
...
+            .iter()
+            .map(|entry| format!("'{}'", entry.raw.replace('\'', r"\'")))
+            .collect();
+        let entries_joined = quoted_entries.join(" ");
+        let formatted_command = template_str.replace("{}", &entries_joined);
+        debug!("Multiple entries command: {:?}", formatted_command);
+        Ok(formatted_command)
+    } else {
+        // Complex braces: use existing template processing
+        debug!("Using complex template processing for: {}", template_str);
```
alexpasmantier, line 124: hmmm I'm not sure I understand this.

The target I had in mind was: say the action command is `nvim {some_template}`. Then we want to format that as `nvim 'R1'separator'R2'separator'R3'` where `separator` is what you already use and `R1`, `R2` and `R3` are the result of applying `some_template` to each of the three entries separately.

alexpasmantier: And unless I'm mistaken what I understand from this is that you're actually using the template to format the resulting concatenated (and quoted) string.

(e.g. `template::format("'e1' 'e2' 'e3'")`)

alexpasmantier: (and what I'm saying above would be more like 
```rs
quote(template::format(e1.output()?)) + separator
+ quote(template::format(e2.output()?)) + etc.
```
)

alexpasmantier: if that makes sense

alexpasmantier: Okay wait a minute I'm myself getting confused now 😅

alexpasmantier: Oh okay I think I understand what's going on.

Yeah I believe the right target is actually using a `Template` like you initially proposed but applying the template to each entry individually **and only then** quote and concatenate with the separator.

I got lost there with `output`, my bad 🙇🏻

alexpasmantier: And that is a bit trickier because it means the template has a special behavior:

```sh
# suppose we're using the text channel where entries look like `file:line_number:...`
# we'd want something like:
nvim +{split:\::1} {split:\::0}

# which would render for a single entry as:
nvim +33 Cargo.toml

# and for multiple entries, well vim isn't a good example because you can't really 
# open multiple files at specified lines that easily but if that did work the result would be:
nvim +33 Cargo.toml +22 README.md +14 CHANGELOG.md

```

alexpasmantier: Any thoughts on this ?

alexpasmantier: Oh and I'm actually pretty sure this can be done (rather) effortlessly by tweaking https://github.com/lalvarezt/television/blob/48190a76e84f9eb7d68323621fbf107c53eebbf2/television/channels/prototypes.rs#L24-L54

EDIT: or by using `MultiTemplate::sections` or something like that 🤔 

EDIT2: would exposing something like `MultiTemplate::format_multiple` make sense?

lalvarezt: uhmm, this is a different use case. Don't know how to deal with that one honestly

Let's call this A
> nvim +{split:\::1} {split:\::0}

Let's call this B
> nvim +33 Cargo.toml

Let's call this C
> nvim +33 Cargo.toml +22 README.md +14 CHANGELOG.md

- If we have a single selection going from A -> B is logical.
- If we have multiple selections it's not evident (at least to me) that A -> C is the logical thing to happen, especially because this is not what happens  with the regular `output`.

A counter example, if I do
`bat {split:\::0} --line-range 0:{split:\::1}`

What would happen if I select multiple items?

Perhaps adding a new notation in string_pipeline could help? To show some intent. Something like
- `nvim {{split:\::1} {split:\::0}}` 
- `nvim ||({split:\::1} {split:\::0})`

This way I would expect that everything that's inside the outer delimiter will repeat for every iteration

This example would look like this then
```
bat {{split:\::0}} --line-range 1:{split:\::1}
# bat file1.txt file2.txt --line-range 1:40 
```

What do you think? We don't have to decide right now, we could leave it like multiple selections and multiple braces it's not supported for the moment 🤷

lalvarezt: being honest after playing with it a bit I don't like how it looks, it adds a lot of complexity for something very niche, at this point it might be better to just create a bash script that does what you want and call that instead

alexpasmantier: Ok you convinced me. Let's merge this for now and see what feedback we get with 0.13.

alexpasmantier (approved):

alexpasmantier:
Happy to merge this when you are 😃

PR #677, opened 2025-08-03
Channels for Kubernetes deployments, pods and services

LeoVen:
## 📺 PR Description

Added Channels for Kubernetes deployments, pods and services

## Checklist


cable/unix/k8s-pods.toml
```diff
@@ -0,0 +1,53 @@
...
+
+[keybindings]
+ctrl-d = "actions:delete"
+ctrl-e = "actions:exec"
+ctrl-l = "actions:logs"
+
+[actions.exec]
+description = "Execute shell inside the selected Pod"
+command = "kubectl exec --stdin --tty -n {split: :0} {split: :1} -- /bin/sh"
+mode = "execute"
```
lalvarezt: Using this with output_mode doesn't make much sense right now. Executed basically exits television and opens the other command so there's no capture involved

cable/unix/k8s-pods.toml
```diff
@@ -0,0 +1,53 @@
...
+output_mode = "capture"
+
+[actions.delete]
+description = "Delete the selected Pod"
+command = "kubectl delete -n {split: :0} pod/{split: :1}"
+mode = "fork"
```
lalvarezt: This mode is not currently implemented, but it will be in the future 
See https://alexpasmantier.github.io/television/docs/Users/channels#channel-specification

cable/unix/k8s-deployments.toml
```diff
@@ -0,0 +1,24 @@
...
+  kubectl get deployments -o go-template --template '{{range .items}}{{.metadata.namespace}} {{.metadata.name}}{{"\n"}}{{end}}'
+  ''',
+  '''
+  kubectl get deployments -o go-template --template '{{range .items}}{{.metadata.namespace}} {{.metadata.name}}{{"\n"}}{{end}}' --all-namespaces
+  ''',
+]
+output = "{split: :1}"
+
+[preview]
+command = "kubectl describe deployments {split: :1} -n {split: :0}"
```
lalvarezt: this is also an option `{0}` is syntactic sugar for `{split: :0}`
```
command = "kubectl describe deployments {1} -n {0}"
```

docs/01-Users/10-community-channels-unix.md
```diff
@@ -363,11 +363,12 @@ description = "A channel to select from git reflog entries"
 requirements = [ "git",]
 
 [source]
-command = "git reflog"
-output = "{split: :0}"
+command = "git reflog --decorate --color=always"
+output = "{strip_ansi|split: :0}"
```
lalvarezt: it's not by much but this is faster.
```
output = "{split: :0|strip_ansi}"
# or "{0|strip_ansi}"
```
No need to do expensive strip_ansi to the whole text when you can do it only to the part you care about

lalvarezt (commented):
here are some notes, if you don't mind

cable/unix/k8s-pods.toml
LeoVen: I knew this was not implemented, but I thought of leaving it there for when the feature is eventually implemented. But if you prefer, I can leave it out or adapt it to `mode = "execute"`.

docs/01-Users/10-community-channels-unix.md
LeoVen: I ran the docs generation and this popped up. I'm not the original author, but if you allow me I can adapt this :)

alexpasmantier: sure let's update that :)

alexpasmantier:
Thanks for the PR!

Also @LeoVen, would you mind updating the screenshot so that it's more compact visually (increasing the font size of your terminal)?

PR #690, opened 2025-08-07
feat: add support for NetBSD installation

0323pin:
## 📺 PR Description

Add NetBSD install instructions

I've just packaged `television` for NetBSD and merged the package into pkgsrc main branch.
Commit log for reference, https://mail-index.netbsd.org/pkgsrc-changes/2025/08/07/msg328025.html

Could you please consider adding this to the supported platforms?
Thanks!

Regards

### DIFF
--- a/docs/01-Users/01-installation.md
+++ b/docs/01-Users/01-installation.md
@@ -73,6 +73,12 @@ sudo dpkg -i tv-$VER-x86_64-unknown-linux-musl.deb
 pixi global install television
 ```
 
+## NetBSD (pkgsrc)
+
+```bash
+pkgin install television
+```
+
 ## Pre-compiled Binary
 
 From the [latest release](https://github.com/alexpasmantier/television/releases/latest) page:


alexpasmantier:
Awesome, thanks for the contribution!

alexpasmantier:
> Could you please consider adding this to the supported platforms?

Did you have something else in mind than adding the installation instructions to the docs?

0323pin:
@alexpasmantier Honestly, adding installation instructions to the docs was my purpose. This makes the platform visible when people visit the website.

Do you have any questions?

I can say that, I will be maintaining the package and try to keep it up-to-date within a reasonable time frame. The only if here depends on the MSRV required. For example, `Cargo.toml` says 1.87, I currently have 1.88 installed but, NetBSD currently has 1.86 as default. This time, it was possible to patch the requirement down to 1.86 and the package built fine. A fellow dev using defaults could confirm this.

You can search packages online, here's the result for `television`, https://pkgsrc.se/textproc/television
This usually takes up to 24h to be visible. Somehow Repology takes longer, one can expect up to 2 days before it shows the correct version there.

Finally, stable package branches are cut every 3 months on NetBSD. This means, `television` will be available to general users after the next release in September. For those users running on "current", _i.e._ rolling packages it's available now. Although, it has to be built from source as we don't provide binary builds for these.
Personally, I don't use releases. I run the development branch of the OS with rolling packages.

If you wonder anything else, just ask and, thank you for considering the addition.

alexpasmantier:
Thanks for the clarification 🙏

Alright let's merge this 🎉

0323pin:
Thank you.

PR #691, opened 2025-08-07
feat(channels): Channel for Justfile recipes

mheuvel-dev:
## 📺 PR Description

Add a channel to select and preview recipes from a Justfile

## Checklist


mheuvel-dev:
I also added a line in my `config.toml`, so that this channel is triggered by `ctrl-t` from command line:
```
"just-recipes" = ["just"]
```

In accordance with the contribution guidelines this line is not added to this PR.

cable/unix/just-recipes.toml
```diff
@@ -0,0 +1,10 @@
+[metadata]
+name = "just-recipes"
+description = "A channel to select recipes from Justfiles"
+requirements = ["just"]
+
+[source]
+command = ["just --summary | tr '[:blank:]' '\n'"]
+
+[preview]
+command = "just -s {}"
```
alexpasmantier, line 10: What do you think about running the selection on completion?
```toml
[metadata]
name = "just-recipes"
description = "A channel to select recipes from Justfiles"
requirements = ["just"]

[source]
command = ["just --summary | tr '[:blank:]' '\n'"]

[preview]
command = "just -s {}"

[keybindings]
enter = "actions:execute-recipe"

[actions.execute-recipe]
description = "Execute a justfile recipe"
command = "just {}"
mode = "execute"
```

alexpasmantier:
Thanks so much for the contribution!

Added a small suggestion, tell me what you think 😊

cable/unix/just-recipes.toml
mheuvel-dev: Thank you for the fast review and the suggestion.

I would prefer a different keybind than `enter`, for example `ctrl-x` for execute, because my usual workflow is using `ctrl-t` for command-line completion. The proposed action is conflicting in that case and may give unintended console output.

alexpasmantier: Sure! That makes sense, let's make the change and merge this :-)

mheuvel-dev: I already made the change.

Thanks again. I like this tool!

alexpasmantier: Ah nice, alright let's merge this 🎉

Thanks 🙏

PR #707, opened 2025-08-20
fix(shell): use $TTY for paste bracketing override

forivall:
## 📺 PR Description

use the $TTY var in the zsh integration so that the paste bracketing is disabled in the correct terminal.

See https://github.com/alexpasmantier/television/pull/512#pullrequestreview-3137810333

## Checklist

- [ ] ~~if this is a new feature, I have added tests to consolidate the feature and prevent regressions~~
- [ ] ~~if this is a bug fix, I have added a test that reproduces the bug~~ N/A (if applicable) N/A
- [ ] ~~I have added a reasonable amount of documentation to the code where appropriate~~


alexpasmantier:
Thanks for the fix 🙏🏻

PR #715, opened 2025-08-27
test: support passing tv binary path

jcgruenhage:
## 📺 PR Description

Add a compile time env var override for the path to the `tv` binary. The tests that run `tv` expect the path to be `./target/debug/tv`, but if you pass a target to cargo, or if you run tests with `--release`, that path doesn't work anymore.

This came up for the FreeBSD people (see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=288915#c5), and also for me while updating the Chimera Linux package.

For an example of how this can be used, see https://github.com/jcgruenhage/cports/blob/a1dd86b5d24de91f1f2b72afa4c371753d074e1d/user/television/template.py#L5-L7.

## Checklist

### DIFF
--- a/tests/cli/special.rs
+++ b/tests/cli/special.rs
@@ -113,7 +113,7 @@ fn test_tv_pipes_correctly() -> io::Result<()> {
         dbg!("Skipping test_tv_pipes_correctly in CI environment");
         return Ok(());
     }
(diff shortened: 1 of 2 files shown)


alexpasmantier:
Thanks for the addition, merging this 🙏🏻

PR #717, opened 2025-09-01
feat(channels): Channel for guix packages

gs-101:
## 📺 PR Description

Adds a cable to search for packages in [Guix](https://guix.gnu.org/). Has an action for spawning a shell containing the chosen package.

<details>

screenshot displaying the Guix cable, a cable for chooosing Guix packages. Only the name of the packages appear. A test search is made for GNU Hello.

</details>

>[!NOTE]
>I thought that it would be a nice idea to have an action that executes the package's binary if it has any **and** the binary matches the package's name.
> I got this to work, but didn't find it quite nice because this wouldn't support passing arguments to the binary. Is there any way to do so?

## Checklist

- [X] my commits **and PR title** follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format


gs-101:
<details>

Screenshot displaying a Hyprland environment with tree windows, one displaying a useless Nix output for television, a second a text editor, and a third showing a browser with this page open.

</details>

Trying to make something for Nix too :v

alexpasmantier:
Thanks for the addition 🙏🏻 

Regarding your question, the only current way of doing that would be to use tv's shell autocomplete feature and enter your arguments manually once you've found your binary.

PR #718, opened 2025-09-01
Default cable dir

RossSmyth:
## 📺 PR Description

Motivation: Packagers can set the environment variable for users so that television works when first downloaded without needing to download cables first.

Allow for `CABLE_DIR` to be set, which becomes the default cable_dir that is looked up. This can be overridden by setting the `--cable-dir` CLI flag. There is some logic to be done so that this isn't a breaking change:

1. If the .config cable directory already exists, use it.
2. If it doesn't exist, check if the environment variable is set
3. If it is set, use it
4. If it is not set, use the .config directory.

This ensures that is `CABLE_DIR` is used in packaging, it will not break users with their own custom cable directories.

## Checklist

### DIFF
--- a/docs/01-Users/09-cli.md
+++ b/docs/01-Users/09-cli.md
@@ -460,7 +460,8 @@ minimum height is ensured (set by default at 15 lines)
 **Purpose**: Uses a custom directory for channel definitions
 
 - **Both Modes**: Same behavior
-- **Default**: `~/.config/tv/cable/` (Linux/macOS) or `%APPDATA%\tv\cable\` (Windows)
+- **Default**: `~/.config/tv/cable/` (Linux/macOS) or `%APPDATA%\tv\cable\` (Windows),
+                or `$CABLE_DIR` if set and `.config/tv/cable` doesn't exist.
 - **Use Case**: Custom channel collections or shared team channels
 
 ### 📚 History Options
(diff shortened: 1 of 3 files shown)


RossSmyth:
I'm not sure the best way to test this. I just deleted my .config cable directory and then ran the tests.

alexpasmantier:
Hi, thanks for taking the time to make this PR.

I like the idea, but am not sure I understand it completely. For instance, would packagers include a default set of cable channels along with setting this env var?

Also for precedence, would something like `cli flag >> env variable >> default location` do the trick ? (I'm willing to remove the cable dir field from the config file since that's primarily a leftover from earlier versions of tv).

RossSmyth:
> I like the idea, but am not sure I understand it completely. For instance, would packagers include a default set of cable channels along with setting this env var?

Yeah, the goal is basically that when you pull Television from a package repo it Just Works without needing to download anything else, but users can also have their own custom cable directory.

alexpasmantier:
> > I like the idea, but am not sure I understand it completely. For instance, would packagers include a default set of cable channels along with setting this env var?
> 
> Yeah, the goal is basically that when you pull Television from a package repo it Just Works without needing to download anything else, but users can also have their own custom cable directory.

I do feel like the `CABLE_DIR` env var is desirable and will definitely add that but maybe what you're after can be achieved by running `tv update-channels` as a post install script?

RossSmyth:
Maybe for some package managers, but I have done some work for Television in Nixpkgs which does not have post-install scripts.

alexpasmantier:
Closing this as the primary motivation was I believe addressed by bffa6e6

PR #719, opened 2025-09-04
feat(channels): add procs channel to manage running processes

Efterklang:
## 📺 PR Description

This channel list running processes via the `ps` command. I've added the following line to my `config.toml` to enable completion for `kill, proc. caffeinate` commands:

```diff
[shell_integration.channel_triggers]
+ "procs" = ["kill", "procs", "caffeinate"]
```

Here is a demo, first I run `sleep 60m`. Then use `tv procs` to search for the process and press `Ctrl-K` to send a `SIGKILL` signal and terminate the process.

https://github.com/user-attachments/assets/98abae48-887b-410a-9bcf-cdc94e840630

## Limitation

- [x] ~~When a command is too long, the remaining part will be truncated. Currently, i have no solution for this.~~

image

## Checklist

### DIFF
--- a/.github/workflows/generate-cable-docs.yml
+++ b/.github/workflows/generate-cable-docs.yml
@@ -8,7 +8,6 @@
 name: Generate Cable Docs
 permissions:
   contents: write
-  pull-requests: write
 on:
   push:
     paths:


cable/unix/procs.toml
```diff
@@ -0,0 +1,21 @@
...
+requirements = ["ps"]
+
+[source]
+command = "ps -e -o pid=,ucomm= | awk '{print $1, $2}'"
+display = "{split: :1}"
+output = "{split: :0}"
+
+[preview]
+# command = "ps -p '{split: :0}' -v"
+command = "ps -p '{split: :0}' -o user,pid,ppid,state,%cpu,%mem,command"
```
CheariX:
```suggestion
command = "ps -p '{split: :0}' -o user,pid,ppid,state,%cpu,%mem,command | fold"
```
You can pipe the preview to `fold` to fix the long lines bug.

Efterklang: it works, thanks for the suggestion.

alexpasmantier:
Awesome! Thanks for the contribution 🙏🏻

PR #720, opened 2025-09-04
fix(channels): deduplicate history command in nu-history channel

Efterklang:
…tory channel

## 📺 PR Description

When I use the nu-history channel to find historical commands, there are many duplicate entries, which are unnecessary for me. Therefore, I tried to modify the channel's source and use uniq to deduplicate commands.

The left one is the origin version and right one is modified version

image

## Checklist


alexpasmantier:
Thanks for the tweak 🙏🏻 
Merging this.

PR #738, opened 2025-10-05
feat(channels): Channel for selecting tv channels

mheuvel-dev:
## 📺 PR Description

Channel for selecting television cable channels.

I would prefer this channel to be builtin. This channel doesn't respect command line flags `--config-file` or `--cable-dir`. Instead it would be nice to allow immediate jump into the remote-control, for example, using `tv <ctrl-t>`. Nevertheless, the channel does the job.  

## Checklist


mheuvel-dev:
I don't know why the UT fails on this channel.

lalvarezt:
isn't this what this flag does? Show the remote directly

```sh
tv --show-remote
```

mheuvel-dev:
It doesn’t have a keybinding (ctrl-t) and there is also no tab completion for channels. The remote also doesn’t show the available keybindings for a channel.

I agree that having those items builtin would be preferred. Meanwhile this channel could be an alternative (together with a binding in the config.toml)

mheuvel-dev:
Please note that the failing test case might show some issue regardless of this PR.

alexpasmantier:
Thanks for the pull request, merging this 🎉

Issue #743, opened 2025-10-14
television Crash Report


radiclestar:
**Description**

Panic occurs when running tv in nushell pipeline and typing "destr" in the tv stdin input box.  I can type longer text example "destyo" and everything works as expected but I get reliable crashes typing "destr"

**Example**

```nu
commandline edit --insert (
  history
  | get command
  | reverse
  | uniq
  | str join (char -i 0)
  | tv --ansi --source-entry-delimiter '\0' --show-preview --preview-command "nu -c '^echo -n `{}` | nu-highlight'"
  | decode utf-8
  | str trim
)
```

**Environment**

- channel stdin

**Expected behavior**

Client not crashing

**Additional context**

name = "television"
operating_system = "NixOS 25.11.0 [64-bit]"
crate_version = "0.13.2"
explanation = """
Panic occurred in file 'television/screen/result_item.rs' at line 159
"""
cause = "range end index 39 out of range for slice of length 38"
method = "Panic"
backtrace = """
   0:     0x55642aaa55bf - human_panic::handle_dump::h921b1290c2a5523c
   1:     0x55642adbb4ac - television::errors::init::{{closure}}::h6f0650bb6b4c4c95
   2:     0x55642ac54d49 - std::panicking::rust_panic_with_hook::h346ebaa42917f592
   3:     0x55642ac76308 - std::panicking::begin_panic_handler::{{closure}}::h0b05e73d9c1fac68
   4:     0x55642ac76269 - std::sys::backtrace::__rust_end_short_backtrace::hc30d33c93b2d1f3b
   5:     0x55642ac545cc - __rustc[8ed47d4163cac052]::rust_begin_unwind
   6:     0x55642a8e931f - core::panicking::panic_fmt::h00861d2bf36d4e75
   7:     0x55642a8e8f26 - core::slice::index::slice_end_index_len_fail::do_panic::runtime::h000868171b7ce2bc
   8:     0x55642a8e8eb9 - core::slice::index::slice_end_index_len_fail::h55f386dbb1d33384
   9:     0x55642adba75e - television::screen::result_item::build_entry_spans::hf4d3a8d0d81189d5
  10:     0x55642ada95a3 - television::draw::draw::h12917e65d7270f5e
  11:     0x55642a9327a5 - television::render::render::{{closure}}::h7286e9b66a06d6cb
  12:     0x55642a95f600 - tokio::runtime::task::raw::poll::hb2de1a0ff7c5a2ff
[...]

PR #747, opened 2025-10-25
fix: build_entry_spans out-of-bounds range

mheuvel-dev:
Fixes https://github.com/alexpasmantier/television/issues/743

## Checklist

### DIFF
--- a/television/screen/result_item.rs
+++ b/television/screen/result_item.rs
@@ -150,7 +150,7 @@ fn build_entry_spans<T: ResultItem + ?Sized>(
     let chars: Vec<char> = entry_name.chars().collect();
     let mut idx = 0;
     for &(start, end) in &match_ranges {
-        let start = start as usize;
+        let start = (start as usize).min(chars.len());
         let end = end as usize;
         if idx < start {
             let text: String = chars[idx..start].iter().collect();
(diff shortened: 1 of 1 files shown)


alexpasmantier:
Thanks for the investigation!
This is a nice safeguard to have, but I wonder why the indices are out of bounds to start with 🤔 Did you get any chance to check that?

mheuvel-dev:
I've just solved problem reported by the crash report.
However, it seems there is something with initializations of stdin that can cause problems, but I couldn't fully finish that investigation.

alexpasmantier:
This does look like an off-by-one error:
```
 cause = "range end index 39 out of range for slice of length 38"
```

mheuvel-dev:
I think the root cause has to do with refreshing the stdin buffers. I also see occasional crashes with subsequent television runs or changing inputs. I just had another crash report where some bad bash input triggered a panic. Next run was ok. Maybe slow hardware or heavy load can be the problem? I will try to investigate further issues related to this but not now.

At least this change prevents crashes and allows local testing of boundaries.

alexpasmantier:
True, I just don't want us to have that bug silenced and forget about it with this fix though...

alexpasmantier:
I believe this ended up being fixed, closing the PR 🙇🏻

Issue #693, opened 2025-08-08
television Crash Report


dryBoneMarrow:
**Description**
`tv update-channels` crashes

**Example**
Run `tv update-channels`

**Expected behavior**
Shouldn't crash

**Environment**

- OS: Arch
- TV version: television 0.12.5
- shell: bash
- terminal: kitty
- config: Default, deleted the old one and used the auto-generated one

**Additional context**
Output:
```
~> tv update-channels
Fetching latest cable channels...
  Discovered channel: alias.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/alias.toml
  Discovered channel: aws-buckets.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/aws-buckets.toml
  Discovered channel: aws-instances.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/aws-instances.toml
  Discovered channel: bash-history.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/bash-history.toml
  Discovered channel: dirs.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/dirs.toml
  Discovered channel: docker-images.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/docker-images.toml
  Discovered channel: dotfiles.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/dotfiles.toml
  Discovered channel: env.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/env.toml
  Discovered channel: files.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/files.toml
  Discovered channel: fish-history.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/fish-history.toml
  Discovered channel: git-branch.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/git-branch.toml
  Discovered channel: git-diff.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/git-diff.toml
  Discovered channel: git-log.toml		download url: https://raw.githubusercontent.com/alexpasmantier/television/main/cable/unix/git-log.toml
[...]

PR #793, opened 2025-11-27
fix(update-channels): channel configs not downloading

hayderhassan:
## 📺 PR Description
Firstly let me just say what an amazing job you've done on this project. I recently switched from neovim to zed and needed a replacement for telescope, which is when I came across television, and personally I prefer this.

## ❗Problem

When I run `tv update-channels`, the logs say that channels were downloaded successfully and there’s no error code. However, nothing actually gets downloaded, even if I use the `--force` flag.

## 🔍 Root Cause

- I realised that the issue is that I haven't set the `$GITHUB_TOKEN` variable, which is used by the `update-channels.sh` script when defining the auth header for the first `curl` command.
- I'm not sure if we're supposed to set this variable but when I removed the auth header the `curl` command worked.
- The `jq` filter causes the subsequent while loop's `$relpath` variable to be in the format `$root/<channel-name>.toml`, e.g. `cable/unix/alias.toml`, and then the `curl` command uses this `$relpath` to form the destination path where the channel's `toml` file will be saved, e.g. `~/.config/television/cable/unix/alias.toml`.
- I've been using `~/.config/television/cable/` as the directory to save channel configurations, and not `~/.config/television/cable/unix/`, but channels can work whether they are saved in `cable/` or in `cable/unix/`.
- If there are two channels with the same name, one saved in `cable/` and one in `cable/unix/`, then the former will take precedence.
- The only issue with saving channels in `cable/unix/`, is that the `channels` channel's preview window uses `bat` to output the content of that channel's file found in the `cable/` directory, and therefore channels saved in `cable/unix/` cannot be previewed in the `channels` channel. That has to be a world record for the number of times the word "channel" was used in a sentence!

## 🎯 Intended Behaviour
The solution depends on what your desired behaviour is, namely:

1. **Should users be required to set a `GITHUB_TOKEN`?**
   - From my understanding the token serves two purposes:
     - Enable access to private repos
     - Allow up to 5000 requests/hour per user, as opposed to the standard 60 requests/hour per IP
   - In this case I think it would be safe allowing unauthenticated requests as it's a public repo and it's unlikely that a user will exceed 60 requests/hour
2. **Should channel configs be saved in `~/.config/television/cable/` or `~/.config/television/cable/unix/`?**
   - The documentation advises to save configs in `cable/`, however the `update-channels.sh` script saves configs in `cable/unix/`.
   - I think it's useful to keep the current behaviour, i.e. when calling `tv update-channels` save the default configs in `cable/unix/` but expect users to add custom channels to `cable/`, which always takes precedence. This way a user could implement their own version of an existing channel, e.g. `cable/alias.toml` and if they run an update then `cable/unix/alias.toml` will be updated, while their custom config will persist. 

## 💡 Proposal
If you agree with the points above then I'd like to suggest this PR as a potential fix:

1. **`scripts/update-channels.sh`**
   - Remove the Authorization header from `curl` to ensure that the script still works without a `GITHUB_TOKEN`
2. **`cable/unix/channels.toml`**
   - Preview the contents of a channel by checking for its config file either directly in `cable/` or any of its subdirectories

## ✅ Outcome

- Users don’t need to set a GitHub token to run `tv update-channels`
- Default configs stay in `cable/unix/`
- User customisations saved in `cable/` will always take precedence and survive updates
- `tv channels` previews a channel regardless of whether its config is saved in `cable/` or `cable/unix/`

---

## Checklist

### DIFF
--- a/cable/unix/channels.toml
+++ b/cable/unix/channels.toml
@@ -7,7 +7,7 @@ requirements = ["tv", "bat"]
 command = ["tv list-channels"]
 
 [preview]
-command = "bat -pn --color always ${XDG_CONFIG_HOME:-$HOME/.config}/television/cable/{}.toml"
+command = "bat -pn --color always ${XDG_CONFIG_HOME:-$HOME/.config}/television/cable/**/{}.toml"
 
 [keybindings]
 enter = "actions:channel-enter"
@@ -16,4 +16,3 @@ enter = "actions:channel-enter"
 description = "Enter a television channel"
 command = "tv {}"
 mode = "execute"
-
--- a/scripts/update-channels.sh
+++ b/scripts/update-channels.sh
@@ -9,10 +9,7 @@ dest="${XDG_CONFIG_HOME:-$HOME/.config}/television"
 
 mkdir -p "$dest"
 
-auth_header=()
-[[ -n "${GITHUB_TOKEN:-}" ]] && auth_header=(-H "Authorization: Bearer $GITHUB_TOKEN")
-
-curl -sS -L "${auth_header[@]}" \
+curl -sS -L --fail \
   "https://api.github.com/repos/$owner/$repo/git/trees/$ref?recursive=1" \
 | jq -r --arg root "$root/" '
     .tree[]
@@ -28,4 +25,3 @@ curl -sS -L "${auth_header[@]}" \
   done
 
 echo "Done -> $dest"
-


alexpasmantier:
Thanks for the kind words and the bug report.

The `tv update-channels` subcommand actually does not use the shell script you're mentioning (which was really just added as a workaround [for certain specific cases](https://github.com/alexpasmantier/television/issues/693)). The code that is used is [here](https://github.com/alexpasmantier/television/blob/main/television/gh.rs). You might want to run the application locally and check the logs (see https://github.com/alexpasmantier/television/blob/main/CONTRIBUTING.md#developing-locally for more info) to try and pinpoint the issue.

Regarding the script, I believe the current code already falls back to an unauthenticated call when `GITHUB_TOKEN` is not defined:
```sh
auth_header=()
[[ -n "${GITHUB_TOKEN:-}" ]] && auth_header=(-H "Authorization: Bearer $GITHUB_TOKEN")
```

Regarding the cable channel (`channels`), I do feel comfortable with the current logic (unique name inside cable/**/, `tv update-channels` writes to `cable/`, etc.) and the `update-channels` command will already check for existing custom entries so as not to overwrite them (the workaround script does overwrite entries but it's really just there for cases where the default command won't work).

This being said, I do agree with the changes to the `channels` channel, which follows the cable dir specification more closely.

Would you be ok dropping the change on `update-channels.sh`, or improving it to avoid overwriting entries so that it's more consistent with the default command?

hayderhassan:
Thanks for the feedback, I’ve now raised two PRs to address the issues:

1. **[fix(channels-preview): preview configs in cable subdirectories](https://github.com/alexpasmantier/television/pull/794)**  
   - Updated `tv channels` so that preview works for config files saved in `cable/` or any of its subdirectories (e.g. `cable/unix/`).  
   - The channel now looks for configs in `${XDG_CONFIG_HOME:-$HOME/.config}/television/cable/**/{}.toml` instead of only `${XDG_CONFIG_HOME:-$HOME/.config}/television/cable/{}.toml`, which aligns with where `tv update-channels` saves configs.

2. **[fix(update-channels): prevent overwriting user configs](https://github.com/alexpasmantier/television/pull/795)**  
   - Added logic to skip downloading a channel config if it already exists locally, unless the `--force` flag is set.  
   - The script also checks the system OS and sets `root="cable/$OS"` instead of the previous hardcoded `root="cable/unix"`, so it now handles both Windows and Unix cables.  
   - For the `curl` call, I removed the `auth_header` array (which was failing under `set -euo pipefail` when `GITHUB_TOKEN` was unset) and replaced it with:  
     ```bash
     curl -sS -L ${GITHUB_TOKEN:+-H "Authorization: Bearer $GITHUB_TOKEN"} …
     ```  
   - After filtering config paths from `jq` and passing them into the loop, each path is checked locally. If it exists, the download is skipped unless `--force` is specified.
   - The script should no longer overwrite user configs by default, but it still allows explicit overwrites when requested.

PR #795, opened 2025-11-28
fix(update-channels): prevent overwriting user configs

hayderhassan:
## 📺 PR Description
This PR prevents the `update-channels.sh` script from overwriting existing user configs unless explicitly requested.   

- Added logic to skip downloading a channel config if it already exists locally, unless the `--force` flag is set.   

- Introduced OS detection to set `root="cable/$OS"` instead of the previous hardcoded `root="cable/unix"`, allowing support for both Windows and Unix cables.   

- Simplified the curl authentication logic by replacing the `auth_header` array (which failed under `set -euo pipefail` when `GITHUB_TOKEN` was unset) with inline conditional expansion:   

  ```bash 

  curl -sS -L ${GITHUB_TOKEN:+-H "Authorization: Bearer $GITHUB_TOKEN"} … 

  ``` 

- After filtering config paths from `jq`, each path is checked locally. If it exists, the download is skipped unless the `--force` flag is set. 

- This ensures user configs are preserved by default, while still allowing explicit overwrites when needed. 

## Checklist

### DIFF
--- a/scripts/update-channels.sh
+++ b/scripts/update-channels.sh
@@ -1,18 +1,26 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+case "$OSTYPE" in
+  linux-gnu* | darwin*) OS="unix" ;;
+  msys* | cygwin*) OS="windows" ;;
+  *) echo "Unsupported OS: $OSTYPE" && exit 1 ;;
+esac
+
 owner="alexpasmantier"
 repo="television"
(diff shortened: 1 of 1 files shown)


alexpasmantier:
Hey, sorry for the late reply on this one.

Thanks for the change 👍🏻

hayderhassan:
No worries

PR #828, opened 2025-12-26
docs: fix command option in README for tv preview

driftcell:
## 📺 PR Description

The command should be `fd -t f . | tv --preview-command 'bat -n --color=always {}'`, I think the `--preview` subcommand is outdated.

## Checklist


alexpasmantier:
Ah, good catch!

Thanks for the fix 👍

PR #830, opened 2025-12-27
refactor(zsh): use the dedicated channel for the zsh history shell integration

### LINKED ISSUE #829: Ability to configure ui.input_bar.position for ctrl+r history search
**Problem**
Right now `ctrl+r` uses `ui.input_bar.position` from main config so there is no way to easily configure it to be in the top for all commands and in the bottom for `ctrl+r` history search only.

**Additional context**
I tried to configure it in `zsh-history` channel and it works for `tv zsh-history` but it doesn't apply to `ctrl+r`. Looks like `ctrl+r` shell integration doesn't use `zsh-history` channel.

Combined with https://github.com/alexpasmantier/television/issues/194 it would make the behavior very similar to fzf. IMHO history search is the last thing where fzf behaves better. Let's fix it please.

alexpasmantier:
fixes #829

### DIFF
--- a/television/utils/shell/completion.zsh
+++ b/television/utils/shell/completion.zsh
@@ -93,7 +93,7 @@ _tv_shell_history() {
 
     local output
 
-    output=$(history -n -1 0 | tv --no-status-bar --input "$current_prompt" --inline $*)
+    output=$(tv zsh-history --no-status-bar --input "$current_prompt" --inline $*)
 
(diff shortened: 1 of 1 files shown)


keturn:
This reverted #379—the use of the shell built-in `history` instead of the channel (which relies only on $HISTFILE) was deliberate.

PR #846, opened 2026-01-08
refactor(nix): flake using flake-parts, crane, and rust-overlay

doprz:
## 📺 PR Description

This PR modernizes the Nix flake infrastructure and adds support for non-flake Nix workflows.
Also thank you for creating such an awesome tool!

Features:
- Shell completions now installed automatically for bash, zsh, fish, and nushell
- Runtime dependencies (bat, fd, jq, just) automatically added to PATH via wrapper
- Improved file filtering using lib.fileset to include only necessary files in build
- Traditional Nix support via new shell.nix and default.nix for non-flake users
- Improved dev shell with all necessary tools

Flake Modernization:
- Migrated from `naersk` to `crane` for improved Rust build support and better caching
- Replaced `nixpkgs-mozilla` with `rust-overlay` for a more reliable and modern Rust toolchain management
- Adopted `flake-parts` for better code organization and maintainability

Misc:
- Updated `.gitignore`
- Updated `README.md`

References:
- https://crane.dev
- https://flake.parts/
- https://github.com/oxalica/rust-overlay

## Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,6 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
     flake-parts.url = "github:hercules-ci/flake-parts";
-    flake-utils.url = "github:numtide/flake-utils";
     crane.url = "github:ipetkov/crane";
     rust-overlay = {
       url = "github:oxalica/rust-overlay";
@@ -64,7 +63,10 @@
             license = lib.licenses.mit;
             platforms = lib.platforms.unix;
             mainProgram = "tv";
-            maintainers = [ lib.maintainers.doprz ];
+            maintainers = [
+              lib.maintainers.doprz
+              "tukanoidd"
+            ];
           };
 
           # Common arguments can be set here to avoid repeating them later
--- a/flake.lock
+++ b/flake.lock
@@ -33,24 +33,6 @@
         "type": "github"
       }
     },
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1731533236,
-        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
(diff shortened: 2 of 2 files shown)


doprz:
@tukanoidd it seems you were the one maintaining the nix infrastructure previously, cc-ing you for continuity here.

tukanoidd:
Ye, will take a look in a bit, was planning to move to crane and rust-overlay too at some point but never found time.

flake.nix
```diff
@@ -1,69 +1,142 @@
...
+            nativeBuildInputs = with pkgs; [
+              makeWrapper
+              installShellFiles
+            ];
+
+            buildInputs =
+              [ ]
+              ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [
+                # Additional darwin specific inputs can be set here
+                pkgs.libiconv
```
tukanoidd, line 93: @doprz  I remember this coming from the example template, do we need it here?

tukanoidd:
Apart from that comment, looks good, will test locally to see if it all works as well

tukanoidd:
Yep, all good, just remove that libiconv thing unless it's actually needed

flake.nix
```diff
@@ -1,69 +1,142 @@
...
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-
-    nixpkgs-mozilla = {
-      url = "github:mozilla/nixpkgs-mozilla";
-      flake = false;
-    };
-
+    flake-parts.url = "github:hercules-ci/flake-parts";
     flake-utils.url = "github:numtide/flake-utils";
```
tukanoidd: prolly should also remove this

doprz: Good catch, thanks!

flake.nix
doprz: The template has unrequired things commented out, however, I think `pkgs.libiconv` is usually required on macOS/Darwin for Rust projects as it helps link to C libraries.

tukanoidd: https://www.gnu.org/software/libiconv/ if it's not the correct one, my bad, but it seems to be just a standalone lib for encodings. I'm not a MacOS user, so I don't know if it's somehow an issue with Rust specifically where it has to require this kind of library for everything, including full utf8 terminal applications

doprz: I'm not on macOS either, I'd say let's leave it for now as it's in the official docs + quite a small dep and if any macOS users can test at a later date if this dep is needed then we can update the flake.

tukanoidd: sure

alexpasmantier:
Thanks so much for the PR, and thanks for the thorough review @tukanoidd

PR #849, opened 2026-01-08
feat(preview): Add word wrapping to the preview panel

xosxos:
I have some niche string visualization use cases where piping to `fold` is not quite enough. Also, for me personally, the default behaviour in the preview panel should be word wrapping.

So, for consideration, here is a PR implementing word wrapping. In the future, if more configuration is available, it could be cleaner to send a reference to `MergedConfig` into `draw_preview_content_block` instead of individual fields.

The test is not very robust, but I think it gets the job done for now! 

Cheers

### DIFF
--- a/television/cli/args.rs
+++ b/television/cli/args.rs
@@ -230,11 +230,13 @@ pub struct Cli {
 
     /// Enables preview panel word wrap.
     ///
-    /// Example: `--word-wrap`
+    /// Example: `--preview-word-wrap`
     #[arg(
-        long = "word-wrap",
+        long,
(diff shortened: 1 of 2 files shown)


xosxos:
Damnit, I have to get my clippy and rustfmt settings sorted out :smile: I'll get back to this later

alexpasmantier:
Took the liberty of fixing these linting/formatting issues and renaming the cli argument to `--preview-word-wrap` to try to stay consistent with existing naming.

Thanks for the PR, merging this!

PR #859, opened 2026-01-13
feat(update-channels): skip channels with unmet requirements by default

louis-thevenet:
## 📺 PR Description

This PR adds a new flag `--skip-unmet-requirements` to skip channels with unmet requirements during channel update.

This is useful to avoid updating or displaying channels that are not usable in the current environment.

## Checklist

### DIFF
--- a/television/cli/args.rs
+++ b/television/cli/args.rs
@@ -676,7 +676,7 @@ pub enum Command {
     /// Downloads the latest collection of channel prototypes from github
     /// and saves them to the local configuration directory.
     UpdateChannels {
-        /// Force update on already existing channels.
+        /// Force update on unsupported and already existing channels.
         #[arg(long, default_value = "false")]
         force: bool,
     },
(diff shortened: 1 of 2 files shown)


alexpasmantier:
This is a great idea. I'm wondering if it even shouldn't be the default behavior. What do you think?

louis-thevenet:
This is what I was thinking as well. If the user does not meet the requirements, then the channel is useless to them. `--force` would allow the old behavior if needed.
If we agree on this, I can implement it =)

alexpasmantier:
> This is what I was thinking as well. If the user does not meet the requirements, then the channel is useless to them. `--force` would allow the old behavior if needed. If we agree on this, I can implement it =)

Agreed

alexpasmantier:
Sorry for the delay, and thanks for updating the PR to make it the default behavior 👍🏻

PR #866, opened 2026-01-22
feat(cli): sort list-channels output alphabetically

simono:
## Checklist


alexpasmantier:
Yup, this makes sense, thanks for the PR 🙏🏻

PR #877, opened 2026-01-25
perf(frecency): avoid string allocations in sort comparisons

lalvarezt:
## 📺 PR Description

Hi there, it's been a while. When testing the new frencency feature I noticed it tanked the performance on the home dir, so I did some digging and noticed some allocations on a hot zone that could be improved

See bench below

```sh
❯ ./bench_frecency.sh
Benchmark 1: before-frecency (ab6bbfc6)
  Time (mean ± σ):     122.1 ms ±  13.0 ms    [User: 259.1 ms, System: 136.3 ms]
  Range (min … max):   106.8 ms … 152.6 ms    50 runs

Benchmark 2: after-frecency (6cfca4de)
  Time (mean ± σ):     159.1 ms ±  15.2 ms    [User: 297.3 ms, System: 121.7 ms]
  Range (min … max):   131.7 ms … 192.0 ms    50 runs

Benchmark 3: after-perf-patch (e4c033ff)
  Time (mean ± σ):      86.7 ms ±  10.0 ms    [User: 233.5 ms, System: 123.3 ms]
  Range (min … max):    72.6 ms …  97.7 ms    50 runs

Summary
  after-perf-patch (e4c033ff) ran
    1.41 ± 0.22 times faster than before-frecency (ab6bbfc6)
    1.84 ± 0.27 times faster than after-frecency (6cfca4de)
```

## Checklist

### DIFF
--- a/television/channels/entry_processor.rs
+++ b/television/channels/entry_processor.rs
@@ -3,6 +3,8 @@ use crate::{
     matcher::{injector::Injector, matched_item::MatchedItem},
 };
 use fast_strip_ansi::strip_ansi_string;
+use nucleo::Utf32Str;
+use std::borrow::Cow;
 
 /// Implementors of this trait define two things:
 /// - how to push lines into the matcher, including any preprocessing steps (e.g. stripping ANSI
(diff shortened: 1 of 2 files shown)


alexpasmantier:
Nicely spotted!
And thanks for the quick fix 🙏🏻

PR #891, opened 2026-01-28
feat(fish): add channel name completion

simono:
## Summary

- Add dynamic completion for channel names in fish shell
- Disables file completion fallback (fixes channels mixing with file names)
- Groups channels with "Channel" label to distinguish from subcommands

## Before

When typing `tv <Tab>`, fish showed file names mixed with subcommands - no channel completion.

## After

Channel names are completed dynamically from `tv list-channels` output, grouped and labeled separately from subcommands.

## Test plan

```fish
tv init fish | source
tv <Tab>
# Should show channels labeled "(Channel)" and subcommands with their descriptions
```

🤖 Generated with [Claude Code](https://claude.ai/code)


alexpasmantier:
Thanks for adding that! 🥳

Issue #901, opened 2026-02-08
Display template does not render ANSI escape codes when ansi = true


josephschmitt:
**Description**

   When `ansi = true` is set, ANSI codes render correctly in the raw source output. But when a `display`
   template is used, ANSI codes are not rendered — the ESC byte appears as a null character and the
   sequence (e.g., `[34m`) shows as literal text.

   **Example**

```toml
[metadata]
name = "ansi-test"

[source]
command = "printf '\\033[34mhello\\033[39m world'"
ansi = true
display = "{}"
```

   With `display = "{}"`: `hello` is NOT colored — ANSI renders as literal text.

   Replace `display` with `output` and it handles ANSI as expected.

   The template `{}` should be a no-op passthrough, so the behavior should be identical.

   **Expected behavior**

   ANSI escape codes should render in `display` template output the same way they render in the raw
   source output.

   **Environment**

   - OS: macOS (Darwin 24.6.0, arm64)
   - TV version: 0.15.2
   - Shell: fish
   - Terminal: ghostty (inside tmux)

   **Additional context**

   The ESC bytes (0x1b) are present in the template output (confirmed via hex dump), so the template is
   passing them through — but television's display renderer doesn't interpret them as ANSI sequences.

alexpasmantier:
Thanks for the report.
This isn't a bug and is documented [here](https://alexpasmantier.github.io/television/user-guide/channels#source). I agree the docs could be more explicit about it though.. 😊 
```toml
[source]
command = "rg . --no-heading --line-number --colors 'match:fg:white' --colors 'path:fg:blue' --color=always"
# display = "[{split:\\::..2}]\t{split:\\::2}"  # what's displayed in the UI (incompatible with `ansi = true`)
output = "{strip_ansi|split:\\::..2}"
ansi = true  # whether the results are ANSI formatted
```

Making the two applicable at once isn't trivial and necessitates tradeoffs both "theoretically" (ANSI styled text can be seen as a state machine and mixing that up with custom text transformations breaks the state) and practically (for example you don't want to have your input matching on escape sequences nor on the raw text but on the transformed ansi-less version of it and that means you need to keep quite a lot of things in memory: the raw string (needed for the preview, the output pattern, etc.), the transformed string without ansi (for the matcher), the transformed string with ansi (or at least some info on where these ansi sequences are), etc.)

Issue #900, opened 2026-02-08
regex_replace template transformation outputs literal template string


josephschmitt:
**Description**

The `regex_replace` template transformation documented at https://alexpasmantier.github.io/television/advanced/template-system/ does not seem to work. Instead of applying the regex replacement, television outputs the raw template string as a literal.

Other transformations (`strip_ansi`, `split`, piping them together) all work correctly. Only `regex_replace` appears broken.

**Example**

Minimal custom channel (`~/.config/television/cable/test.toml`):

```toml
[metadata]
name = "test"

[source]
command = "echo 'hello world'"
output = "{regex_replace:\\s+:_}"
```

Open with `tv test`, select an entry.

I also tried these variations, all with the same result:
- `{strip_ansi|split: :1|regex_replace:^~:@}`
- `{regex_replace:^~:$HOME}`
- `{regex_replace:hello:goodbye}`

**Expected behavior**

`regex_replace` should apply the regex substitution to the entry. For example, `{regex_replace:\\s+:_}` on `hello world` should output `hello_world`.

**Actual behavior**

The literal template string is output instead (e.g., `{regex_replace:\\s+:_}`).

**Environment**

- OS: macOS (Darwin 24.6.0, arm64)
- TV version: 0.15.2
- Shell: fish
- Terminal: ghostty (inside tmux)
- Config: default `~/.config/television/config.toml`
- Channel: custom channel (shown above)

**Additional context**

I was trying to use `regex_replace` in the `output` field of a custom channel to expand `~` back to `$HOME` for paths that are display-shortened. The workaround for now is to avoid `regex_replace` and handle the transformation externally.

lalvarezt:
Hi, creator of string_pipeline here. Try

`replace:s/PATTERN/REPLACEMENT`

josephschmitt:
> Hi, creator of string_pipeline here. Try
> 
> `replace:s/PATTERN/REPLACEMENT`

@lalvarezt that seems to be a no-op (vs just printing out the whole template). here’s my test toml I tried:

```toml
[metadata]
name = "test"

[source]
command = "echo 'hello world'"
output = "{replace:s/\\s/_}"
```

When I run `tv test` it outputs the unchanged `hello world` instead of the expected `hello_world`. I even tried with a simpler pure character replacement of `output = "{replace:s/h/_}"` and still I get no replacement.

lalvarezt:
Hi there, 

I made a typo on the previous message the correct syntax is

[...]

PR #902, opened 2026-02-08
feat(cable): add pj channel for project directory finding

josephschmitt:
## 📺 PR Description

Adds a new cable channel for [`pj`](https://github.com/josephschmitt/pj), a fast project directory finder that searches your filesystem for project directories.
image

Works with monorepos by deeply looking for nested projects, too:
image

- Two source modes: default and `--no-nested`
- Preview using `ls`
- Actions: expand path, open in `$EDITOR`, connect to tmux session, open new tmux tab, clear cache
- Keybindings for all actions

Two of the actions require `tmux` to work. Given they’re optional and not the primary (nor even secondary) actions, I figured it was ok to not list `tmux` as a requirement. Let me know if you disagree.

## Checklist

### DIFF
--- a/cable/unix/pj.toml
+++ b/cable/unix/pj.toml
@@ -1,0 +1,48 @@
+[metadata]
+name = "pj"
+description = "Fast project directory finder that searches your filesystem for project directories"
+requirements = ["pj"]
+
+[source]
+command = [
+  "pj --icons --ansi --shorten",
+  "pj --icons --ansi --shorten --no-nested"
+]
+ansi = true
+
+[preview]
+command = "ls -la --color=always \"$(echo {strip_ansi|split: :1} | sed 's|^~|$HOME|')\""
+
+[keybindings]
+enter = "actions:expand"
+ctrl-e = "actions:edit"
+ctrl-t = "actions:connect"
+ctrl-n = "actions:new_tab"
+ctrl-r = ["actions:clear_cache", "reload_source"]
+
+[actions.expand]
+description = "Expand to absolute path"
+command = "echo {strip_ansi|split: :1} | sed 's|^~|$HOME|'"
+mode = "execute"
+
+[actions.edit]
+description = "Open in $EDITOR"
+command = "$EDITOR \"$(echo {strip_ansi|split: :1} | sed 's|^~|$HOME|')\""
+mode = "execute"
+
+# Requires tmux
+[actions.connect]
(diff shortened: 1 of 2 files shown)


josephschmitt:
I wanted to use the native `{regex_replace}` for the tilde expansion, but it doesn’t seem to be working, so I used `sed` instead. Opened an issue regarding `regex_replace` here: https://github.com/alexpasmantier/television/issues/900

josephschmitt:
> I wanted to use the native `{regex_replace}` for the tilde expansion, but it doesn’t seem to be working, so I used `sed` instead. Opened an issue regarding `regex_replace` here: https://github.com/alexpasmantier/television/issues/900

Turns out the docs were just outdated. However it also turns out that the replace syntax can't read the `$HOME` env var, so `sed` remains the best path forward here.

alexpasmantier:
Cool!

A better solution in my opinion might be to make `pj` output resolved absolute paths by default (or behind a flag).

josephschmitt:
> Cool!
> 
> A better solution in my opinion might be to make `pj` output resolved absolute paths by default (or behind a flag).

@alexpasmantier so it does by default, the `--shorten` flag is what returns `~` instead of absolute paths. 

What I was struggling to get working (which you can see via some of the issues I opened) is to get `display` to show colored ansi icons AND the shortened home directory but then `output` to display the fully resolved path.

At first I tried using plain `pj --icons --ansi` which would just give me the absolute paths. But when I used the `display` template syntax to then replace `$HOME` with `~`, the ANSI [coloring on the icons broke](https://github.com/alexpasmantier/television/issues/901).

So what I ended up with instead was to have `command` use the results from `pj --icons --ansi --shorten` and then replace them in `output` since that does seem to work well with ANSI coloring.

I’m open to suggestions, I really would much rather do it how you said. But displaying tilde for the home directory instead of the full path is such a nicer UX and cleans up the list so much that this felt worth it

josephschmitt:
@alexpasmantier Lmk if you have any other feedback or concerns, happy to address

alexpasmantier:
Sorry for the delay, would you mind adding a small screenshot of the channel for the documentation? 😊

josephschmitt:
@alexpasmantier I have a couple of screenshots in the PR. Should they go somewhere in the code itself? Is there an assets folder? Or maybe a command I can run to generate the docs for this? happy to add!

Issue #910, opened 2026-02-12
Fish <enter> won't work with ${ENV}


r4v3n6101:
**Description**
Any action that involves opening a file via `${EDITOR}` will fail in case the `fish` is set as shell. It's caused by unsupported `${}` by `fish` (instead needed `{EDITOR}).

**Example**
Any file opening with `fish`. It may be `files`, `todo-comments` and so on channels.

**Expected behavior**
Calling `/bin/sh` maybe. 

**Environment**

- OS: MacOS
- TV version: 0.15.2
- shell: fish
- terminal: kitty
- **The channel you are using**: files

**Additional context**

```shell
fish: ${ is not a valid variable in fish.
${EDITOR:-vim} 'Cargo.lock'
 ^
```

alexpasmantier:
Thanks for the report. Should be fixed by #921

PR #911, opened 2026-02-12
fix: enhance shell command execution for POSIX-style parameter expansion

guangdawang:


r4v3n6101:
Fixes https://github.com/alexpasmantier/television/issues/910

PR #925, opened 2026-03-02
fix: reset tick counter when channel stops to ensure immediate rendering

rucnyz:
## Summary

Reset the tick counter when a channel transitions from running to stopped, so results from slower source commands are rendered immediately instead of being delayed by up to ~500ms.

## Problem

Any cable channel whose source command takes longer than ~80ms to produce output will experience a noticeable rendering delay. The first batch of results appears ~500ms late (on my laptop), even though the data is already available in the matcher.

### How to reproduce

1. Create a cable channel with a source command that takes >100ms (e.g., a command that queries a database, scans files, or calls an API before outputting results)
2. Launch it with `tv <channel>`
3. Observe that results don't appear until ~500-600ms after launch, despite the source command finishing in ~150-250ms

For comparison, a fast source command (<80ms) renders results almost instantly (~140ms).

### Root cause

`should_render()` uses a tick-based throttling scheme:

1. **`FIRST_TICKS_TO_RENDER` (10 ticks)**: Every tick renders — the "fast-render window"
2. **`RENDERING_INTERVAL_FAST` (3 ticks)**: Every 3rd tick renders while `channel.running()` is true
3. **`RENDERING_INTERVAL` (25 ticks)**: Every 25th tick renders otherwise

```
  fn should_render(&self, action: &Action) -> bool {
      (self.ticks < FIRST_TICKS_TO_RENDER          // first 10 ticks: always render
          || self.ticks.is_multiple_of(RENDERING_INTERVAL)  // then every N ticks
          || (self.channel.running()
              && self.ticks.is_multiple_of(RENDERING_INTERVAL_FAST))  // faster while
  running
          || matches!(action, Action::AddInputChar(_) | ...)  // user input
      )
  }
```

The fast-render window is intended to cover ~200ms (10 ticks × 20ms tick rate). However, each `Action::Render` returned from `update()` is fed back through `update()` by the event loop, also incrementing `self.ticks`. This tick cascade means the 10-tick window is consumed in only **~80ms of wall-clock time**, not 200ms.

In `update()`:
```
  self.ticks += 1;

  Ok(if self.should_render(action) {
      Some(Action::Render)
  } else {
      None
  })
```

When a source command takes >80ms:
- The fast-render window has already closed (ticks ≥ 10)
- The source command finishes → `channel.running()` becomes false
- `RENDERING_INTERVAL_FAST` no longer applies (requires `running() == true`)
- The only remaining render opportunity is `ticks % 25 == 0`, which is **~500ms away**

### Tick-level trace (source command finishes at ~150ms)

**Before fix** — first render with results at **622ms**:
```
tick=9   t=80ms   total=0     running=true   should_render=true   ← last tick in fast window
tick=10  t=140ms  total=0     running=true   should_render=false  ← window closed
tick=11  t=140ms  total=0     running=false  should_render=false  ← source exited
tick=14  t=141ms  total=1050  matched=1050   should_render=false  ← data ready, NO render
  ... (ticks 15–24: should_render=false, data just sitting there)
tick=25  t=622ms  should_render=true                              ← finally renders
```

**After fix** — first render with results at **241ms**:
```
tick=15  t=181ms  total=0     running=true   should_render=true
  [source command exits, EOF received]
tick=1   t=221ms  total=1051  running=false  should_render=true   ← ticks RESET, window reopened
tick=2   t=241ms  total=1051  should_render=true                  ← immediate rendering continues
```

**Fast source command (<80ms)** — unaffected, renders at **141ms**:
```
tick=9   t=100ms  total=1050  running=false  should_render=true   ← data arrives within window
  [first render with results at 141ms]
```

## Solution

Add a `was_running: bool` field to `Television`. After `channel.tick()` in `update()`, detect the running → stopped transition and reset `self.ticks = 0`. This reopens the fast-render window exactly when new results become available.

The check is placed between `channel.tick()` and `update_results_picker_state()` so the results picker is populated with fresh data in the same cycle as the reset.

## Test plan

- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo build --release` succeeds
- [x] Tested with cable channel source commands taking ~150-250ms — first render improved from ~622ms to ~321ms
- [x] Verified fast source commands (<80ms) are unaffected (~141ms)
- [x] No behavioral change when channel stays running (e.g., streaming/watch commands)

### DIFF
--- a/television/television.rs
+++ b/television/television.rs
@@ -372,6 +372,7 @@ impl Television {
             frecency_config,
             false, // stdin only applies to the initial channel
         );
+        self.was_running = true;
         self.channel.load();
     }
 


alexpasmantier:
Thanks a lot for the investigation and the detailed PR description.
And thanks for fixing this 👍🏻

PR #926, opened 2026-03-02
chore: modify nightly source to https format

atlarator:
## 📺 PR Description

It's not a good idea to assume everyone has uploaded an ssh-publickey to alexpasmantier's account.

## Checklist


alexpasmantier:
That does make sense, thanks for fixing that 👍🏻

PR #942, opened 2026-03-20
fix: select-1 with prefilled input

### LINKED ISSUE #940: `--select-1` should(?) respect input
**Problem**
`--select-1` doesn't take `--input` into account.

As mentioned in this comment to the initial feature:
https://github.com/alexpasmantier/television/issues/440#issuecomment-2797331295

I have a helper function in nushell:
```nu
def --env cr [query?: string] {
  let path = tv git-repos --select-1 --input $"($query)"
  cd $path
}
```
which enables me to quickly change repo, however i'd often write something 

**Feature**
Either expanding or creating a new `--select-1` related flag, so one can easier utilize tv as a "fallback" on a fuzzy finding use case.

**Examples**

`tv git-repos --select-1 --input "te"`
Works as currently

`tv git-repos --select-1 --input "television"`
would instantly select this repo, similar to `--take-1`.

**Additional context**
In my cases i'd probably like it to select the one as soon as there is only 1 result left, even if i write it - so i dont have to press `enter` when i have one result - but that is less crucial, and could be opt in.

lalvarezt:
## 📺 PR Description

This PR closes #940

```bash
echo aaaa\nwwww\neeee\nrrrr\ntttt\nyyyy | tv -- --select-1 --input 'a'
```

Running this command now automatically selects `aaaa` entry. The `--select-1` command now does so after `--input` is applied

## Checklist

### DIFF
--- a/television/app.rs
+++ b/television/app.rs
@@ -374,10 +374,10 @@ impl App {
 
             if self.television.merged_config.select_1
                 && !self.television.channel.running()
-                && self.television.channel.total_count() == 1
+                && self.television.channel.result_count() == 1
             {
                 // If `self.select_1` is true, the channel is not running, and there is
-                // only one entry available, automatically select the first entry.
+                // only one matched entry available, automatically select it.
                 if let Some(outcome) = self.maybe_select_1() {
                     action_outcome = outcome;
                 }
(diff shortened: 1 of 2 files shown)


alexpasmantier:
Nice, thanks for taking the time to fix that @lalvarezt ❤️

PR #946, opened 2026-03-21
fix(config): bat cannot print file paths that starts with '-' on windows

LeeSF03:
## 📺 PR Description

when trying to print content of file paths that starts with '-' on windows, bat puts out this error:
error: unexpected argument '-1' found

  tip: to pass '-1' as a value, use '-- -1'

Usage: bat.exe [OPTIONS] [FILE]...
       bat.exe <COMMAND>

For more information, try '--help'.

### DIFF
--- a/cable/windows/dotfiles.toml
+++ b/cable/windows/dotfiles.toml
@@ -8,4 +8,4 @@ command = "fd -t f . \"$env:USERPROFILE\\AppData\\Roaming\\\""
 
 
 [preview]
-command = "bat -n --color=always '{}'"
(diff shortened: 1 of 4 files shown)


LeeSF03:
I'm using a windows system and don't use a unix system, so I can't confirm if this problem exists on unix and unix-based systems

alexpasmantier:
Ha! Never thought a path might even start with '-' 😅 
That must be a pain for **a lot** of cli operations, not just when using tv...
Anyway, the PR does fix it without introducing any regression I can think of so thanks for taking the time to do that 👍🏻

PR #949, opened 2026-03-21
fix(previewer): prevent panic when scroll exceeds content lines length

CodeMarco05:
## 📺 PR Description

Fixes a panic in `previewer/state.rs` that occurred when `scroll` exceeded the actual number of lines in `content.lines`.
The root cause was that `for_render_context` used `total_lines` to compute slice bounds, but `total_lines` and `content.lines.len()` can be out of sync (e.g. while a new preview is still loading). This caused:

range start index 4906 out of range for slice of length 1

**Fix:** clamp `scroll` to the real content length and derive `num_lines` from `content.lines.len()` instead of `total_lines`.

Fixes panic reported via human-panic in `television/previewer/state.rs:60`.

## Checklist

- [ x] my commits **and PR title** follow the [conventional commits]

### DIFF
--- a/television/previewer/state.rs
+++ b/television/previewer/state.rs
@@ -52,10 +52,9 @@ impl PreviewState {
     // What if we did it only when the preview content or scroll changes?
     pub fn for_render_context(&self, height: usize) -> Self {
         // PERF: this allocates every time
-        let scroll = self.scroll as usize;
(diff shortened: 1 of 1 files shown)


CodeMarco05:
Thanks for the fast approval. Love the commitment in the project <3

PR #963, opened 2026-03-22
fix(tests): replace unsafe env var manipulation with CommandBuilder::env

qaqland:
The previous implementation used `std::env::set_var` and `std::env::remove_var` within `unsafe` blocks, which could cause SIGSEGV due to race conditions when modifying the process environment while other threads are reading it.

```
     Running tests/command_line.rs (target/debug/deps/command_line-502b1fa4ccd52715)
running 112 tests
error: test failed, to rerun pass `--test command_line` Caused by:
  process didn't exit successfully: `/builds/qaqland/aports/testing/television/src/television-0.15.3/target/debug/deps/command_line-502b1fa4ccd52715` (signal: 11, SIGSEGV: invalid memory reference)
```

## Checklist

### DIFF
--- a/tests/cli/errors.rs
+++ b/tests/cli/errors.rs
@@ -132,31 +132,30 @@ fn test_watch_and_selection_flags_conflict_errors() {
 /// Tests that --inline conflicts with --height.
 #[test]
 fn test_inline_and_height_conflict_errors() {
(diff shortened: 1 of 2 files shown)


alexpasmantier:
Thanks for spotting that, and thanks for the fix 👍🏻

PR #964, opened 2026-03-22
fix(tests): use --take-1 in test_cli_working_directory_override

qaqland:
--take-1 waits for loading to complete before selecting the first entry, making it more reliable than --take-1-fast which may race with data loading.

Also switch to assert_raw_output_contains_with_timeout for more reliable output verification.

Ref https://github.com/alexpasmantier/television/commit/6fbfc23fc41a9da342679d2573b1c40792d23906

## Checklist


alexpasmantier:
Thanks again for taking care of these flaky tests, much appreciated 👍🏻

PR #965, opened 2026-03-22
feat(cable): add jujutsu (jj) channels

delafthi:
## 📺 PR Description

Adds six cable channels for [jujutsu](https://jj-vcs.github.io/jj/) (`jj`), covering the most common day-to-day workflows:

| Channel | Description |
|---------|-------------|
| `jj-bookmark` | Browse and manage bookmarks (create, edit, delete, set, forget, track) |
| `jj-diff` | Browse files changed in the working copy (restore, edit, squash, absorb) |
| `jj-files` | Browse all files tracked in the working copy (edit, restore) |
| `jj-log` | Browse and act on change log entries (edit, new, insert before/after, diff, squash, abandon, describe, rebase) |
| `jj-op-log` | Browse the operation log — jj's equivalent of `git reflog` (restore, undo) |
| `jj-remotes` | List and manage git remotes (fetch, remove) |

Screenshots are included for most channels.

## Checklist


alexpasmantier:
Sorry for the delay!

These look great! One tiny nit I might have is that <kbd>ctrl-x</kbd> is already bound to toggling the actions menu. This means a user that is used to that and starts opening these new channels might hit ctrl-x and get a bad surprise 🙃 

I will need to work on the keybindings a bit sooner or later since the default really aren't very satisfactory, but I'd rather we not create confusion in the meantime.
Would you mind changing that `ctrl-x` to something else you feel is right?

alexpasmantier:
Awesome, thanks for the quick reaction 👍🏻

delafthi:
> I will need to work on the keybindings a bit sooner or later since the default really aren't very satisfactory, but I'd rather we not create confusion in the meantime. 

It would be really great to support upper case bindings (or `shift-<key>`) as well as mentioned in #966. This would unlock a whole lot of possibilities (for terminals that support it). Maybe at some point even support for key sequences.

PR #967, opened 2026-03-22
feat(cable): add macOS diskutil and launchd-services channels

delafthi:
## 📺 PR Description

Add two macOS-specific channels modelled after their Linux counterparts (`mounts` and `systemd-units`):

- **diskutil**: lists mounted filesystems via `df`, previews with `df + ls`, actions for cd, open in Finder, `diskutil info`, eject, and verify
- **launchd-services**: lists launchd services (portrait layout), previews with `launchctl print` across `gui/user/system` domains, actions for start, stop, restart, enable, disable, plist (opens in `$EDITOR`), and log

## Checklist


alexpasmantier:
Same thing here, this looks great but would you mind changing that ctrl-x keybinding to something else? 🙏🏻

PR #993, opened 2026-03-28
feat: native image and PDF preview support

abaj8494:
## Summary

- Add inline image rendering in the preview pane using ratatui-image, with automatic detection of terminal graphics protocol (Kitty/iTerm2/Sixel) from environment variables and font size from ioctl
- Add native PDF preview by converting the first page via pdftoppm and rendering through the same pipeline
- Support remote image/PDF URLs with download-to-cache for cable channels, EXIF orientation, and a "Loading..." indicator while previews generate
- Add `TERM=dumb` to bat-using cable channels to prevent OSC 10/11 escape sequence leaks

### Image rendering
- Introduces `PreviewContent` enum (`Text | Image`) to the preview system
- Detects image files by extension and decodes via the `image` crate with EXIF orientation support
- Renders using `ratatui-image` `StatefulImage` with per-entry protocol state caching
- Graphics protocol detected from env vars (`KITTY_WINDOW_ID`, `TERM_PROGRAM`, `LC_TERMINAL`) — no terminal escape queries, so no conflicts with crossterm's event reader
- Font cell size obtained via `TIOCGWINSZ` ioctl for accurate image scaling

### PDF rendering
- Converts first page of PDFs to JPEG via `pdftoppm` (from poppler-utils)
- Cached by content hash to avoid re-rendering

### Cable channel support
- Remote URLs (http/https) are downloaded to a cache directory and previewed natively
- Uses `entry.output()` for path detection to correctly handle tab-separated cable channel entries
- Switches `files.toml` preview from `bat` to `tv-preview` for broader format support

### New dependencies
- `ratatui-image` — terminal image rendering
- `image` — image decoding with EXIF support
- `rustix` — `tcgetwinsize` ioctl for terminal pixel dimensions

## Test plan
- [ ] Run `tv files` in a Kitty terminal and select an image file — should render inline
- [ ] Select a PDF file — should render first page as image
- [ ] Select a text file — should render normally with syntax highlighting
- [ ] Run in iTerm2/WezTerm — should auto-detect protocol and render images
- [ ] Run in a terminal without graphics support — should fall back gracefully
- [ ] Verify no escape sequence leaks appear in the search bar
- [ ] Test with a cable channel that uses tab-separated output format


abaj8494:
there is a `Loading...` whilst the preview is Loading; much like [lf](https://github.com/gokcehan/lf) provides previews through Kitty

PR #998, opened 2026-03-28
fix(actions): fix execute/fork actions when tv runs under shell integration

### LINKED ISSUE #997: Edit action in `files` channel does not work if run via shell integration
**Description**
`tv` has a nice feature of [opening selected files in editor](https://github.com/alexpasmantier/television/blob/f937ebb15e65aa118fdbb34329c03d1768d992f8/cable/unix/files.toml#L15).
That works fine unless `tv` is run with a shell shortcut (`ctrl-t` by default).
In that case the whole terminal window freezes and can only be killed from that point.

**Example**
Good behaviour: just run `tv` and hit `f12` (default) to open in `$EDITOR`.
Bad behaviour: `source $(tv init fish)` (or `bash`), hit `ctrl-t` and then hit `f12` -- terminal freezes.
This can also be reproduce with a simpler `set selected $(tv)` line (with following `f12`).
The bug is reproduced both in fish and bash.

I checked this behaviour against `fzf` to exclude `fish`-specific problems. `fzf` works correctly in a similar scenario:
```fish
set result (fzf --bind 'f12:execute(nvim {})') # nvim opens normally.
```
 

**Expected behavior**
Editor and other programs should open normally regardless of how `tv` was started.

**Environment**

- OS: NixOS
- TV version: 0.15.4
- shell: fish
- terminal: kitty

lalvarezt:
## 📺 PR Description

This PR closes #997 

It reattaches external action stdio to `/dev/tty` on Unix before exec or spawn, so actions launched from shell integration use the real terminal instead of inheriting command-substitution pipes

## Checklist

### DIFF
--- a/tests/cli/external_actions.rs
+++ b/tests/cli/external_actions.rs
@@ -143,3 +143,204 @@ fn test_external_action_thebatman_with_f8() {
         "expected cat output to contain 'Copyright (c)', got:\n{output}"
     );
 }
+
+/// Verifies that execute-mode actions attach their stdout to the
+/// controlling tty, not to tv's stdout. We run tv inside a bash script
+/// that captures its stdout (`out=$(tv ...)`); with the fix, the action's
+/// `TTY_OK` output lands on the terminal (visible in the scrollback),
+/// and the captured `$out` stays empty.
+#[test]
+fn test_execute_action_uses_tty_when_stdout_is_captured() {
+    let pt = phantom();
+
+    let tempdir = TempDir::new().unwrap();
+    let cable_dir = tempdir.path().join("captured_stdout");
+    fs::create_dir_all(&cable_dir).unwrap();
+
+    let files_toml_content = r#"
+[metadata]
+name = "files"
+description = "A channel to select files and directories"
+requirements = ["fd", "bat"]
+
+[source]
+command = ["fd -t f", "fd -t f -H"]
+
+[preview]
+command = "bat -n --color=always '{}'"
+env = { BAT_THEME = "ansi" }
+
+[keybindings]
+shortcut = "f1"
+f12 = "actions:ttycheck"
+
+[actions.ttycheck]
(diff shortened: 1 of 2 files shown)


tests/cli/external_actions.rs
```diff
@@ -151,3 +152,158 @@ mode = "execute"
     // Check that the process has finished
     PtyTester::assert_exit_ok(&mut child, DEFAULT_DELAY);
 }
+
+#[test]
+fn test_execute_action_uses_tty_when_stdout_is_captured() {
+    let mut tester = PtyTester::new();
+
+    let cable_dir = TempDir::new().unwrap().path().join("captured_stdout");
+    fs::create_dir_all(&cable_dir).unwrap();
```
alexpasmantier: I think the `cable_dir` never gets cleaned up here because the tempdir is dropped immediately.
You probably need to:
```rs
let temp_dir = ...;
let cable_dir = ...;
```

tests/cli/external_actions.rs
```diff
@@ -151,3 +152,158 @@ mode = "execute"
...
+        output
+    );
+}
+
+#[test]
+fn test_fork_action_uses_tty_when_stdout_is_captured() {
+    let mut tester = PtyTester::new();
+
+    let cable_dir =
+        TempDir::new().unwrap().path().join("captured_stdout_fork");
```
alexpasmantier: same thing here

alexpasmantier (commented):
Great job investigating this 👍🏻 
Thanks for the fix.
Left a minor comment, but otherwise lgtm!

PS: sorry for the delay

tests/cli/external_actions.rs
lalvarezt: you're right, fixed

alexpasmantier:
Thanks for the fix 😁

PR #1000, opened 2026-03-29
fix(channels): deduplicate bash-history entries

mslugyfw:
## 📺 PR Description

The `bash-history` channel currently displays all entries from `HISTFILE`, including duplicate commands that accumulate over time.

This PR adds `| awk '!seen[$0]++'` to the source command to filter out duplicates, keeping only the most recent occurrence of each command.

## Checklist


alexpasmantier:
Thanks for the quality of life improvement 🙌🏻

PR #1008, opened 2026-03-31
feat(channels): flush candidate batches by size or interval

Bryley:
## 📺 PR Description

This PR improves candidate streaming responsiveness by flushing batches not only when they reach the large batch size threshold, but also when a fixed time interval elapses.

### What changed
- Added a time-based flush interval (`200ms`) for candidate ingestion.
- Kept the existing size-based flush behavior (`BATCH_SIZE`) intact.
- Applied this behavior consistently to both command-based candidate loading and stdin candidate loading.

### Why
Previously, output was primarily flushed on large batch boundaries (or at end-of-stream), which could delay visible updates for slower or incremental sources.  
This change improves perceived responsiveness by delivering partial results earlier while preserving throughput behavior for large streams.
### Notes
- This is intended as a UX/performance improvement for streaming and ad-hoc command scenarios.

## Checklist

- [X] my commits **and PR title** follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format
- [X] if this is a new feature, I have added tests to consolidate the feature and prevent regressions
- [X] I have added a reasonable amount of documentation to the code where appropriate

### DIFF
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -19,8 +19,11 @@ use std::process::Stdio;
 use std::sync::Arc;
 use std::sync::atomic::AtomicBool;
 use std::time::Duration;
-use tokio::io::{AsyncBufReadExt, BufReader};
 use tokio::process::Command as TokioCommand;
+use tokio::{
+    io::{AsyncBufReadExt, BufReader},
+    time::Instant,
+};
 use tracing::debug;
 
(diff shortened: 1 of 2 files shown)


alexpasmantier:
That's a great idea and the implementation looks neat. Thanks for the contribution 👍🏻 

fixes #1035

PR #1028, opened 2026-04-09
feat: add custom source command names for multi-source channels

### LINKED ISSUE #1025: feat: add support to add custom name for source commands (when more than one source command used)
**Problem**
When using several source commands it's hard to distinguish which source is active now, since I see only string "Results" and source tabs indicator. I have 3 source commands for example for Claude sessions: Active, Archived, All and can't see exactly which source I'm currently on.

**Feature**
Add abitily to specify custom name for source command so, when that source command selected custom name shown instead of "Results"

nick4eva:
## 📺 PR Description

Add an optional `command_name` field to the `[source]` section of cable channel TOML definitions. When multiple source commands are defined, these names are displayed in the results panel header instead of the generic "Results" label, making it easier to identify which source is currently active.

Example usage:
```
[source]
command = ["ls", "ls -a", "ls -la"]
command_name = ["Default", "Hidden", "Detailed"]
```
Closes #1025

named_source_commands_1

named_source_commands_2

named_source_commands_3

## Checklist

### DIFF
--- a/benches/main/previewer.rs
+++ b/benches/main/previewer.rs
@@ -1,5 +1,4 @@
 use criterion::{Criterion, criterion_group};
-use rustc_hash::FxHashMap;
 use std::hint::black_box;
 use television::{
     channels::{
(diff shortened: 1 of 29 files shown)


alexpasmantier:
Thanks for suggesting the idea and taking the time to put it in writing.
Took the liberty of refactoring the PR slightly to make this a bit more robust by having a structured field for command specs instead of maintaining two parallel arrays for the commands themselves and the names.
Merging this 🎉

PR #1030, opened 2026-04-09
feat(tui): enables the kitty keyboard-protocol and enable support for uppercase keybindings

### LINKED ISSUE #966: Control + upper case keybindings
**Problem**
`ctrl-G` doesn't behave differently from `ctrl-g`, and `ctrl-shift-g` doesn't appear to bind at all. Is this intentional, a known limitation, or am I missing something in the configuration?

**Feature**
Support for modifier + uppercase key combinations like `ctrl-G` / `ctrl-shift-g` as distinct bindings from their lowercase counterparts.

**Examples**
```toml
"ctrl-g" = "some_action"
"ctrl-G" = "some_other_action"  # currently treated the same as ctrl-g
```

**Additional context**
From reading the code it looks like two things may be at play: the key string is fully lowercased before parsing, so `ctrl-G` becomes `ctrl-g` before anything else happens.
https://github.com/alexpasmantier/television/blob/f937ebb15e65aa118fdbb34329c03d1768d992f8/television/config/keybindings.rs#L137

And modifier combinations like `ctrl+shift` don't seem to have a corresponding `Key` variant, so they'd fall through silently. 
https://github.com/alexpasmantier/television/blob/f937ebb15e65aa118fdbb34329c03d1768d992f8/television/event.rs#L383-L388

If this is just not supported yet, it would be a nice addition — `ctrl-G` being distinct from `ctrl-g` opens up a good chunk of extra keyspace for channels with many custom bindings.

delafthi:
## 📺 PR Description

This PR checks if the terminal supports the [kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/) and enables it if supported. It also adds support for defining uppercase bindings, such as `Ctrl-D` or `Ctrl-Shift-d`.

However, on terminals that don't support the kitty keyboard protocol (like tmux; let's hope [tmux PR #4912](https://github.com/tmux/tmux/pull/4912) gets merged), the shift key information is not propagated to `tv`. This means that if bindings for both `ctrl-d` and `Ctrl-D` are defined, only the `ctrl-d` action will be executed, since the shifted binding cannot be distinguished from the unshifted one.

The feature was tested in Ghostty.

## Checklist

Resolves #966

### DIFF
--- a/television/event.rs
+++ b/television/event.rs
@@ -380,12 +380,19 @@ pub fn convert_raw_event_to_key(event: KeyEvent) -> Key {
             KeyModifiers::ALT => Key::AltSpace,
             _ => Key::Null,
         },
-        Char(c) => match event.modifiers {
-            KeyModifiers::NONE | KeyModifiers::SHIFT => Key::Char(c),
-            KeyModifiers::CONTROL => Key::Ctrl(c),
-            KeyModifiers::ALT => Key::Alt(c),
-            _ => Key::Null,
-        },
+        Char(c) => {
+            let c = if event.modifiers.contains(KeyModifiers::SHIFT) {
+                c.to_uppercase().next().unwrap_or(c)
+            } else {
+                c
+            };
+            match event.modifiers - KeyModifiers::SHIFT {
+                KeyModifiers::NONE => Key::Char(c),
+                KeyModifiers::CONTROL => Key::Ctrl(c),
+                KeyModifiers::ALT => Key::Alt(c),
+                _ => Key::Null,
+            }
+        }
         _ => Key::Null,
     }
 }
@@ -430,7 +437,7 @@ mod tests {
             kind: KeyEventKind::Press,
             state: KeyEventState::NONE,
         };
-        assert_eq!(convert_raw_event_to_key(event), Key::Char('a'));
+        assert_eq!(convert_raw_event_to_key(event), Key::Char('A'));
 
         let event = KeyEvent {
             code: KeyCode::Char(' '),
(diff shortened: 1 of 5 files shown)


television/event.rs
```diff
@@ -380,12 +380,19 @@ pub fn convert_raw_event_to_key(event: KeyEvent) -> Key {
...
+            } else {
+                c
+            };
+            match event.modifiers - KeyModifiers::SHIFT {
+                KeyModifiers::NONE => Key::Char(c),
+                KeyModifiers::CONTROL => Key::Ctrl(c),
+                KeyModifiers::ALT => Key::Alt(c),
+                _ => Key::Null,
+            }
+        }
```
delafthi, line 395: When a shifted key is pressed, it is reported as:

```
2026-04-09T20:48:46.857697Z TRACE television/event.rs:328: Raw event: KeyEvent { code: Char('d'), modifiers: KeyModifiers(SHIFT | CONTROL), kind: Press, state: KeyEventState(0x0) }
```

Here the shift modifier is first applied to the character, then the other modifiers are processed. This ensures that shifted keys work correctly on characters with all currently supported modifiers.

delafthi:
I'm struggling with the tests. I'm not sure why some of them are breaking. Probably, because some of them have timing issue due to the additional check for kitty keyboard protocol support.

delafthi:
@alexpasmantier, should I keep this ready for merging? This is  just a basic implementation, and I opened the pull request to gather some feedback. I haven’t spent much time figuring out why the tests are failing and thought you could point me into the right direction.

If you’re interested, you can also ping me to update this PR when you believe this feature is ready to be merged.

alexpasmantier:
Just been really busy these past weeks, sorry about that.
Let's leave this open for now, I'll have a look when I get a bit of free time :-)

delafthi:
> Just been really busy these past weeks, sorry about that.

No pressure 😉

alexpasmantier:
Back to this
Awesome PR, thanks 🙏🏻 
Took the liberty of tweaking a few things:
- querying the terminal for support when creating the TUI (instead of each time we enter it, e.g. after a suspend) and after entering the alternate screen
- updated a couple of docs
- refactored key parsing a bit so that it's less implicit

alexpasmantier (approved):

PR #1032, opened 2026-04-10
feat(cable): add zellij-sessions channel

Vexo413:
## 📺 PR Description

Add a zellij-sessions channel for searching through the Zellij multiplexer's sessions, similar to the tmux-sessions channel.

## Checklist

### DIFF
--- a/cable/unix/zellij-sessions.toml
+++ b/cable/unix/zellij-sessions.toml
@@ -1,0 +1,27 @@
+[metadata]
+name = "zellij-sessions"
+description = "List and manage zellij sessions"
+requirements = ["zellij"]
+
+[source]
+command = "zellij list-sessions | sed 's/\\x1b\\[[0-9;]*m//g'"
+display = "{}"
+output = "{split: :0}"
+
+[preview]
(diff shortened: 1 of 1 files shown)


cable/unix/zellij-sessions.toml
```diff
@@ -0,0 +1,27 @@
+[metadata]
+name = "zellij-sessions"
+description = "List and manage zellij sessions"
+requirements = ["zellij"]
+
+[source]
+command = "zellij list-sessions | sed 's/\\x1b\\[[0-9;]*m//g'"
```
alexpasmantier, line 7: it looks like this is trying to strip ansi escape codes from zellij's output. 

[Reading zellij's code](https://github.com/zellij-org/zellij/blob/main/zellij-utils/src/cli.rs#L292) (couldn't find anything in the docs for some reason), it seems they have a `--no-formatting` option you could use to avoid having to do that.

and as a last resort if the above doesn't work you can also use `display = "{strip_ansi}"`

alexpasmantier:
Thanks for the PR, looks mostly good to me except for that unnecessary sed expression I mentioned above.

PR #1036, opened 2026-04-13
feat(channels): add configurable batch size

### LINKED ISSUE #1035: Allow configurable batch_size per channel
The candidate batch size is hardcoded at 10000 entries. This works well for
fast local commands, but channels backed by slow sources (paginated APIs,
remote queries) become unresponsive: users stare at an empty picker until the
entire first batch fills up (or, more likely, all pages are
loaded).

A `batch_size` field on the source config would let channel authors
tune this per channel:

```toml
[source]
command = "my-slow-api --list"
batch_size = 100
```

The default stays at 10000 so existing channels are unaffected.

I have this implemented and can open a PR.

seflue:
Closes #1035

## 📺 PR Description

Channels can now set `batch_size` in their `[source]` config to control
how many entries accumulate before flushing to the matcher. The default
remains 10,000. This makes channels backed by slow or paginated sources
responsive immediately instead of blocking until the full batch fills.

```toml
[source]
command = "my-slow-api --list"
batch_size = 100
```

- `batch_size` is a new field on `SourceSpec`, deserialized with a
  serde default so all existing channel configs stay unchanged
- Threaded through `MergedConfig` → `ChannelKind::new()` →
  `load_candidates()` / `load_stdin_candidates()`
- `DEFAULT_BATCH_SIZE` is `pub` for use in tests and benchmarks

**To test:** create a channel with a slow source command (e.g.
`for i in $(seq 1 50); do echo $i; sleep 0.1; done`) and set
`batch_size = 5`. Items should appear in the picker as soon as the
first batch fills instead of waiting for the full output. Without the
option set, behavior is identical to `main`.

## Checklist


television/channels/channel.rs
```diff
@@ -251,11 +256,8 @@ impl<P: EntryProcessor> Channel<P> {
 }
 
 const DEFAULT_LINE_BUFFER_SIZE: usize = 256;
-// Batch size for pushing candidates to the injector
```
lalvarezt, line 254: don't remove this line, it has useful information

television/channels/channel.rs
```diff
@@ -29,6 +29,7 @@ pub struct Channel<P: EntryProcessor> {
     pub source_command: CommandSpec,
     pub source_entry_delimiter: Option<char>,
     pub source_output: Option<Template>,
+    batch_size: usize,
```
lalvarezt: for consistency make this public

television/channels/channel.rs
seflue: Restored the comments. Sorry about that.

television/channels/channel.rs
seflue: Done.

lalvarezt (approved):

alexpasmantier:
I think the more correct fix to this is PR https://github.com/alexpasmantier/television/pull/1008

Closing this, thanks for taking the time to submit a PR anyway, and thanks @lalvarezt for the reviews! 🙏🏻

PR #1048, opened 2026-04-19
fix(cable): remove redundant quoting around {} in action commands

### LINKED ISSUE #976: Default dirs channel cd action breaks on paths with spaces (double-quoting)
## Description

The default `dirs` cable channel's `cd` action fails when selecting a directory with spaces in its name. This is caused by a double-quoting issue: the action command template manually wraps `{}` in single quotes, but `format_command` in `utils/command.rs` also auto-quotes `{}` entries in action commands (introduced in PR #675).

## Steps to reproduce

1. Have a directory with a space in the name, e.g. `~/Cornwall AI/`
2. Run `tv dirs`
3. Select the directory and trigger the `cd` action

## Expected behaviour

Television should `cd` into the directory.

## Actual behaviour

zsh reports: `zsh:cd:1: string not in pwd: Cornwall`

## Cause

The default `cable/unix/dirs.toml` has:

```toml
[actions.cd]
command = "cd '{}' && $SHELL"
```

Since `format_command` already auto-quotes `{}` in action commands, the resulting command after substitution becomes:

```
cd ''Cornwall AI'' && $SHELL
```

zsh interprets `''` as an empty string, then sees `Cornwall` and `AI` as two separate arguments. The two-argument form of `cd` in zsh means "replace first string with second in `$PWD`", hence the error.

## Fix

Remove the manual single quotes from the action command in the default `dirs.toml`:

```toml
command = "cd {} && $SHELL"
```

Note: the `preview.command` using `'{}' ` is fine because preview commands go through `Template::format()` which does simple string replacement without auto-quoting.

## Environment

- television 0.15.4
- macOS Darwin 25.3.0 (aarch64)
- zsh 5.9

Mic92:
Since `format_command()` already wraps each entry in single quotes when substituting a bare `{}` placeholder, having `'{}'` in the channel config results in `''entry''`, which the shell parses as an unquoted string. This breaks any action on entries containing spaces, e.g. in the `downloads`, `dirs`, `files` and `zoxide` channels:

```
rm: cannot remove '/home/user/Downloads/Ethereum': No such file or directory
rm: cannot remove 'Flakes.pdf': No such file or directory
```

Preview commands are left untouched since they go through `Template::format` which does not auto-quote.

Fixes #976

### DIFF
--- a/cable/unix/aws-profiles.toml
+++ b/cable/unix/aws-profiles.toml
@@ -17,5 +17,5 @@ enter = "actions:export"
 
 [actions.export]
 description = "Export the selected profile (sets AWS_PROFILE)"
-command = "export AWS_PROFILE='{}' && echo \"AWS_PROFILE set to '{}'\" && $SHELL"
+command = "export AWS_PROFILE={} && echo \"AWS_PROFILE set to {}\" && $SHELL"
 mode = "execute"
--- a/cable/unix/brew-packages.toml
+++ b/cable/unix/brew-packages.toml
@@ -21,10 +21,10 @@ ctrl-d = "actions:uninstall"
 
 [actions.upgrade]
 description = "Upgrade the selected package"
-command = "brew upgrade '{}'"
+command = "brew upgrade {}"
 mode = "execute"
 
 [actions.uninstall]
 description = "Uninstall the selected package"
-command = "brew uninstall '{}'"
+command = "brew uninstall {}"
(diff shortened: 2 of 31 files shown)


alexpasmantier:
Thanks for spotting that and thanks for the PR 👍🏻 

Ideally we shouldn't have that special case of automatic quoting since it's kind of counter intuitive. Will look into it.

PR #1054, opened 2026-04-29
fix(cli): prioritize registered channel over path when argument matches both

### LINKED ISSUE #1043: Channel argument treated as path
**Description**
The CLI documentation indicates that the first argument should be a channel name and the second a directory. However, if you have a channel name that happens to have the same name as a subdir of the current directory, tv will go against the documentation and treat the first argument as a dir, then it will run the default command, typically `files`

**Example**
I have the following channel for `npins`

```toml
# npins.toml
[metadata]
description = "A channel to select npins pins for update, freeze, or unfreeze"
name = "npins"
requirements = ["npins", "jq"]

[preview]
command = "jq '.pins[\"{}\"]' npins/sources.json"

[source]
command = "jq -r '.pins | keys[]' npins/sources.json"
```

When I run `tv npins` in my project root with an `npins` subdir, it selects the `files` channel and shows me the files in `npins`

**Expected behavior**
I would expect it would select the `npins` channel with no directory argument.

**Environment**

- OS: NixOS
- TV version: 0.15.4
- shell: fish
- terminal: wezterm
```toml
# config.toml
[ui]
theme = "catppuccin-mocha-mauve"
use_nerd_font_icons = true
```

**Additional context**
This is likely a regression introduced in [e76a3df](https://github.com/alexpasmantier/television/commit/e76a3df#diff-35f3e20b36ec15bd14afc52ae6d83d8b6245ef4f125a1bd1c2108db67fcc62c5R93) in `cli/mod.rs`. Previously it checked for the existence of a channel but that logic got flipped and it now checks if the first argument is a directory and only falls back to treating it as a channel if it isn't.

wucm667:
Fixes #1043

## Problem

When a channel name (e.g., `npins`) coincides with a subdirectory in the current working directory, the CLI incorrectly treated the argument as a path and fell back to the default `files` channel.

For example, running `tv npins` in a project root with an `npins/` subdirectory would show files inside `npins/` instead of opening the `npins` channel.

This is a regression introduced in e76a3df (`refactor(cable)!: cable format redesign`), where the logic was changed from "check if it's a registered channel first" to "check if it's a path first".

## Fix

Restore the pre-e76a3df behavior in `post_process()`:
1. Load the cable directory to get all registered channel names
2. Check if the argument is a registered channel first
3. Only treat it as a working directory if it is NOT a registered channel AND the path exists
4. If it's neither a registered channel nor an existing path, exit with an unknown channel error

## Changes

| File | Change |
|------|--------|
| `television/cli/mod.rs` | Load cable in `post_process`; restore channel-first priority in the channel/path match block |

### DIFF
--- a/television/cli/mod.rs
+++ b/television/cli/mod.rs
@@ -1,13 +1,13 @@
 use crate::{
     action::{Action, Actions},
-    cable::Cable,
+    cable::{CABLE_DIR_NAME, Cable, load_cable},
     channels::prototypes::{ChannelPrototype, Template},
     cli::args::{Cli, Command},
     config::{
         Keybindings, get_config_dir, get_data_dir, merge_keybindings,
         ui::{BorderType, Padding},
     },
-    errors::cli_parsing_error_exit,
+    errors::{cli_parsing_error_exit, unknown_channel_exit},
     event::Key,
     screen::layout::{InputPosition, Orientation},
     utils::paths::expand_tilde,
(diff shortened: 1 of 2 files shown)


delafthi:
This is a real bummer for me. I can't use the new `XDG_PROJECTS_DIR` (per default `=~/Projects`) with my custom `tv projects` command to fuzzy match and launch the selected project session. I would highly appreciate if this could land in the next release.

alexpasmantier:
Tanks for taking the time to fix this.
Had a couple of concerns about the implementation which I took the liberty of refactoring and pushing into one last commit on top of the PR.
Merging 🎉

alexpasmantier (approved):

Issue #1004, opened 2026-03-29
Windows: shell = "bash" resolves to WSL bash instead of Git Bash


NikiforovAll:
## Problem

On Windows with both Git Bash and WSL installed, setting `shell = "bash"` in `config.toml` causes `tv.exe` to resolve `bash` to `C:\Windows\System32\bash.exe` (WSL bash stub) instead of Git Bash (`C:\Program Files\Git\usr\bin\bash.exe`).

This happens because Windows `CreateProcessW` searches `System32` **before** `PATH` directories. Even though Git Bash appears first in `PATH` (verified via `where bash`), the WSL stub in System32 wins.

As a result, commands like `fd` and `bat` fail with:
```
/bin/bash: line 1: fd: command not found
/bin/bash: line 1: bat: command not found
```

## Expected behavior

`tv` should use Git Bash when `shell = "bash"` is configured and Git Bash is installed.

## Proposed solution

Support full executable paths in the `shell` config option, e.g.:

```toml
shell = "C:/Program Files/Git/usr/bin/bash.exe"
```

Currently the `Shell` enum only accepts named variants (`bash`, `zsh`, `fish`, `powershell`, `cmd`, `nu`) via `Shell::executable()`, which returns bare command names. Supporting a full path (or a `shell_path` option) would let Windows users disambiguate between multiple `bash.exe` installations.

## Workaround

Setting `shell = "powershell"` works since PowerShell resolves commands via the standard Windows `PATH`.

## Environment

- Windows 11
- television 0.15.4
- Git Bash (Git for Windows) installed
- WSL installed (`C:\Windows\System32\bash.exe` present)
- `fd` and `bat` installed via winget and on PATH

NikiforovAll:
<img width="3836" height="2111" alt="Image" src="https://github.com/user-attachments/assets/67496670-4c27-4eea-b0d5-e9e65fe1d0a9" />

hossam-houssien:
same as here #987

NikiforovAll:
will close as duplicated, thank you

Issue #987, opened 2026-03-27
television Crash Report - "0.15.4" - win 11


hossam-houssien:
**Description**
# Ruuning `tv` ver "0.15.4" fail on cmd, pwsh or bash shell "win 11"
### using cmd or pwsh shell
**Example**
```shell
> tv
Well, this is embarrassing.

television had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "C:\Users\<user>\AppData\Local\Temp\report-3de719a5-5fa6-465f-a890-7506c847a3da.toml". Submit an issue or email with the subject of "television Crash Report" and include the report as an attachment.

- Homepage: https://github.com/alexpasmantier/television
- Authors: Alexandre Pasmantier <alex.pasmant@gmail.com>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
```
>this behaviour is the same using windows terminal and waveterm.

### using bash shell MSYS2 or Cygwin "win 11" it start the tui but :
on left pan it show this error:
/bin/bash: line 1: fd: command not found
or the right pan show this:
/bin/bash: line 1: fd: command not found
/bin/bash: line 1: bat: command not found

>this behaviour is the same using windows terminal and waveterm.

alsthough fd and bat are in the $PATH
```shell
$ fd --version
fd 10.2.0
$ bat --version
bat 0.25.0 (25f4f96)
```
check uploaded pics
**Expected behavior**
should list files in current directory

**Environment**

- OS: Microsoft Windows [Version 10.0.22631.5624]
- TV version:  television 0.15.4
- shell: cmd,pwsh,bash
- terminal: Windows Terminal Preview: 1.24.2682.0, wave terminal: 0.14.3 (202603122334)
- **Your configuration file (`~/.config/television/config.toml`)**
- default - uploaded
- **The channel you are using** (if applicable and different from the default)
n/a
**Additional context**
when trying :
```shell
cat file|tv 
```
this works fine
**files**
file : `report-3de719a5-5fa6-465f-a890-7506c847a3da.toml`
```log
name = "television"
operating_system = "Windows 10.0.22631 (Windows 11 Professional) [64-bit]"
crate_version = "0.15.4"
explanation = '''
Panic occurred in file 'D:\hossam\bin\rust\registry\src\index.crates.io-1949cf8c6b5b557f\television-0.15.4\television\channels\channel.rs' at line 281
'''
[...]

PR #1057, opened 2026-05-08
Fix windows shell

hossam-houssien:
## 📺 PR Description

issue #987  and #1004  
panic caused by std::process::Command on Windows throwing program not found
the shell names were missing the .exe extension on Windows.
also implemented a default to `pwsh.exe` if available, gracefully falling back to `powershell.exe`.

## Checklist
- [x] I have added a test case in issue #987 that reproduces the bug


Cargo.lock
```diff
@@ -2570,7 +2570,7 @@ dependencies = [
 
 [[package]]
 name = "television"
-version = "0.15.6"
+version = "0.15.6-hos-fix"
```
alexpasmantier: I don't think you want to include this in the PR

television/utils/shell.rs
```diff
@@ -134,12 +134,60 @@ impl Shell {
...
+                        {
+                            "pwsh.exe"
+                        } else {
+                            "powershell.exe"
+                        }
+                    })
+                } else {
+                    "pwsh"
+                }
+            }
```
alexpasmantier, line 190: To avoid all that branching, let's just duplicate the `executable` function and have one for windows and one for unix

alexpasmantier (changes requested):
Thanks for taking the time to fix this. The solution looks fine to me, just left a couple of comments I think we should address before merging.

HeyItsGilbert:
I'm not sure specifying the extension is needed. https://doc.rust-lang.org/std/process/struct.Command.html#platform-specific-behavior

television/utils/shell.rs
HeyItsGilbert: Would a using the which crate and a match be better here?
```suggestion
            Shell::Bash => "bash",
            Shell::Zsh => "zsh",
            Shell::Fish => "fish",
            Shell::Cmd => "cmd",
            Shell::Nu => "nu",
            Shell::Psh => {
                match which("pwsh") {
                    Ok(path) => "pwsh",
                    Err(_) => "powershell.exe",
                }
            }
```

Cargo.lock
hossam-houssien: you are right , I though i remove it

television/utils/shell.rs
hossam-houssien: @alexpasmantier it still branching anyway, but if using `std::process::Command` as @HeyItsGilbert mentioned, this would cut the branch,
@HeyItsGilbert  `std::process::Command` wasn't used anywhere, only used in my update, and you're right it dosen't require the `.exe` extension,however your update still won't work (I guess you're asuming that crate `utils::shell` uses `std::process::Command`, it's not)

PR #1059, opened 2026-05-11
fix(update-channels): Use `GITHUB_TOKEN` to avoid request rate limit

cometkim:
## 📺 PR Description

GitHub may reject requests with 429. Requests within an organization may be rejected more often because they share the same network client IP.

This change gives a way to isolate requests by `GITHUB_TOKEN` when available from the shell environment variable. e.g.

```bash
GITHUB_TOKEN="$(gh auth token)" tv update-channels
```

Just like the shell script version does here:
https://github.com/alexpasmantier/television/blob/f3cb4c4/scripts/update-channels.sh#L23

## Checklist

### DIFF
--- a/television/gh.rs
+++ b/television/gh.rs
@@ -41,7 +41,7 @@ fn make_gh_request(url: &str) -> Result<RequestBuilder<WithoutBody>> {
 
     let headers = request.headers_mut().unwrap();
     if let Ok(token) = std::env::var("GITHUB_TOKEN") {
-        debug!("Use GITHUB_TOKEN from environment variables");
+        debug!("Using GITHUB_TOKEN environment variable");
         headers.insert("Authorization", HeaderValue::try_from(token)?);
     }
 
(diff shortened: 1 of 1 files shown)


television/gh.rs
```diff
@@ -36,14 +36,25 @@ enum NodeType {
 const GITHUB_API_BASE_URL: &str =
     "https://api.github.com/repos/alexpasmantier/television/contents/";
 
+fn make_gh_request(url: &str) -> Result<RequestBuilder<WithoutBody>> {
+    let mut request = get(url).header("User-Agent", "television-client");
+
+    let headers = request.headers_mut().unwrap();
+    if let Ok(token) = std::env::var("GITHUB_TOKEN") {
+        debug!("Use GITHUB_TOKEN from environment variables");
```
alexpasmantier: nit:
```suggestion
        debug!("Using GITHUB_TOKEN environment variable");
```

alexpasmantier:
Thanks for the contribution, LGTM 👍🏻

PR #1066, opened 2026-05-13
feat: support per-shell binary path overrides in config

### LINKED ISSUE #1004: Windows: shell = "bash" resolves to WSL bash instead of Git Bash

### LINKED ISSUE #1065: Add Pwsh vs PowerShell Support
**Problem**
The current deployment assumes PowerShell is the name of the binary. This is fine if you want to launch "Windows PowerShell" (aka 5.1). This doesn't work for the newer version of PowerShell.

**Feature**
Add a new `Shell::Pwsh` which resolves to `pwsh` which is the name of the executable on all OS's. The goal would be to support pwsh which is the newer version of PowerShell that is cross platform.

I've done this previously in starship https://github.com/starship/starship/pull/5478

**Examples**
One or more code examples that shows the feature in action.

**Additional context**
In #1057 there is some fallback setup. This would be to offer to distinguish between the two.

HeyItsGilbert:
## Summary

Fixes #1004
Fixes #1065 

On Windows with both Git Bash and WSL installed, `shell = "bash"` causes `tv.exe` to resolve `bash` to `C:\Windows\System32\bash.exe` (the WSL stub) instead of Git Bash, because `CreateProcessW` searches `System32` before `PATH` directories.

This PR adds a new `shell_binaries` map in `config.toml` that lets users specify exact executable paths per shell variant, so the correct binary is always invoked.

## Usage

```toml
[shell_binaries]
bash = "C:/Program Files/Git/usr/bin/bash.exe"
powershell = "C:/Program Files/PowerShell/7/pwsh.exe"
```

Per-channel overrides are also supported — set `shell_binaries` inside a channel's `[source]` or `[preview]` block. Channel-level entries take priority over the global map.

## Changes

- `Shell` — added `Eq` derive so it can be used as a `HashMap` key
- `AppConfig` — new `shell_binaries: HashMap<Shell, String>` field
- `CommandSpec` — new `shell_binaries: HashMap<Shell, String>` field (per-channel override)
- `shell_command()` — accepts the map and looks up the active shell variant, falling back to the default binary name when not present
- `layers.rs` — global `shell_binaries` entries are merged as fallback into both source and preview `CommandSpec`s
- New tests covering deserialization and binary resolution logic

### DIFF
--- a/.config/config.toml
+++ b/.config/config.toml
@@ -24,6 +24,10 @@ default_channel = "files"
 # If not specified, the shell is detected from the environment ($SHELL on Unix).
 # Channel-specific shell settings override this global setting.
 # shell = "bash"
+# Binary to use for a given shell. Useful on Windows when `bash` resolves to the
+# WSL stub in System32 instead of Git Bash.
+# [shell_binaries]
+# bash = "C:/Program Files/Git/usr/bin/bash.exe"
 # History settings
 # ---------------
 # Maximum number of entries to keep in the global history (default: 100)
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -285,6 +285,7 @@ pub async fn load_candidates<P: EntryProcessor>(
         command.interactive,
         &command.env,
         command.shell,
+        &command.shell_binaries,
     );
     std_command.stdout(Stdio::piped()).stderr(Stdio::piped());
     let mut child = TokioCommand::from(std_command)
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -354,7 +354,6 @@ pub async fn load_candidates<P: EntryProcessor>(
         command.interactive,
         &command.env,
         command.shell,
-        &command.shell_binaries,
     );
     std_command.stdout(Stdio::piped()).stderr(Stdio::piped());
     let mut child = TokioCommand::from(std_command)
--- a/television/channels/channel.rs
+++ b/television/channels/channel.rs
@@ -354,6 +354,7 @@ pub async fn load_candidates<P: EntryProcessor>(
         command.interactive,
         &command.env,
         command.shell,
+        &command.shell_binaries,
     );
     std_command.stdout(Stdio::piped()).stderr(Stdio::piped());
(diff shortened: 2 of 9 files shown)


HeyItsGilbert:
@alexpasmantier anything you need from me?

alexpasmantier:
Hey, thanks so much for the PR and sorry for the delay.
Took the liberty of refactoring things a bit so that the code is simpler and all shell invocations benefit from the overrides automatically. Also dropped the per-channel feature which I thought seemed premature.

alexpasmantier:
Let's merge this 🎉

alexpasmantier (approved):

PR #1071, opened 2026-05-17
Fix: specify shell="bash" for distrobox-list channel

jinliu:
Because fish doesn't like the preview command.


alexpasmantier (approved):
Thanks!

PR #1107, opened 2026-07-08
feat: swap Nucleo for Frizbee

saghen:
## 📺 PR Description

This PR migrates from [Nucleo](https://github.com/helix-editor/nucleo) to [Frizbee](https://github.com/saghen/frizbee) which significantly improves matching performance (~10x in parallel) and allows for typo resistance in the future. Also, Frizbee always sorts on index which makes results more stable than Nucleo (try typing while ingesting data in the linux repo on the `text` channel). 

There's 3 main parts to the PR: `Matcher` rewrite, `Action::MatcherUpdated` event and plumbing

`Matcher`: Runs Frizbee in a dedicated worker thread and receives items from the Channel via an `Injector` similar to the Nucleo approach, but batched. Calls `Notify` when a new snapshot is available.
`Action::MatcherUpdated`: New `Action` which allows the `Matcher` to wake the `EventLoop` when a new match `Snapshot` is available

A couple features are still missing:
- [x] `Frecency`: ~~Currently, if two items are in the list and one has higher frecency, it will beat the other match regardless of its score. This could be implemented as-is, but a slightly more efficient (and perhaps more natural) would be to add the frecency score to the `Match` score. This would let us re-use the existing radix sort inside of Frizbee.~~
- [x] `Cancellation`: ~~When a new pattern comes in, Frizbee will finish matching against the current haystack before restarting matching on the new pattern. Ideally we would cancel when a new pattern comes in, but given the speed, this doesn't seem strictly necessary. I only noticed it on the linux repo with the `text` channel~~ Decided against adding this as the current behavior seems fine
- [x] ~~Reverse matching order~~ Seems to be unused

Frizbee issues (not blocking):
- Weird indices on `matcher::new(Snapshot::` on the `text` channel on this repo

Multi-pattern and substring/exact/prefix/suffix matching are implemented (`foo !^bar`). However, they're only on `main` so I've pointed to the `frizbee` git for now. I'll make a 0.11 release before marking this PR as ready.

Performance wise, Frizbee is almost able to keep up with my keystrokes when testing on the linux repo while Nucleo takes ~10s and noticeably stutters when `pattern.len() == 0`. Ingestion speed seems to be slightly slower, likely due to Frizbee locking the `Store` while matching (preventing items from being added), but I don't mind this trade-off given the implementation is relatively simple.

Related #19


alexpasmantier:
Only had the time to check out the PR and do a couple of tests locally so far but it looks very promising 😊

saghen:
I think the PR is in a good spot now! Couple questions for ya:

- Would you be interested in switching `frencecy` to be a bonus on the `Match.score`? This may make results more natural, and would improve performance during sorting. I haven't measured current performance though.
- How do you want to expose typo resistance to the user? In blink.cmp, we allow `max_typos = keyword.len() / 4` but this may be a bit too much for a fuzzy finder, not sure.

lalvarezt:
Hi there, this is looking great. Sorry for never coming back to you in the comment so long ago 😅, been out of topic for some time. My two cents


> - Would you be interested in switching `frencecy` to be a bonus on the `Match.score`? This may make results more natural, and would improve performance during sorting. I haven't measured current performance though.

I think this would be the best, the closer we are to the hot path the better. One of the things were I struggled on the initial implementation of frecency was the need to wait until the matching was done to post-process the results and include it there (that or vendoring nucleo)


> - How do you want to expose typo resistance to the user? In blink.cmp, we allow `max_typos = keyword.len() / 4` but this may be a bit too much for a fuzzy finder, not sure.

Don't know about this one, fff uses this formula clamped between 2 and 6 so I'd say that would be a good start https://github.com/dmtrKovalenko/fff/blob/44a5b259570730a4236ecbf06673d43ef7b2263e/crates/fff-core/src/file_picker.rs#L957

alexpasmantier:
> Would you be interested in switching frencecy to be a bonus on the Match.score? This may make results more natural, and would improve performance during sorting. I haven't measured current performance though.

yes definitely. any plans on how you might tackle this?

> How do you want to expose typo resistance to the user? In blink.cmp, we allow max_typos = keyword.len() / 4 but this may be a bit too much for a fuzzy finder, not sure.

I actually feel like that's a pretty sound choice. Let's keep that heuristic and see how it feels while dogfooding.

alexpasmantier:
I've actually been (silently) iterating quite a lot on this and have now reached a point I think I'm comfortable with and that (from my manual benchmarks) seems really fast both for ingestion and matching.

The TLDR goes something like this:
- made injectors send batches over a channel so that they never contend on the shared store => ingestion speed is faster than the current release (and it makes the matcher faster too)
- the big one: matching is now made by progressive chunks of doubling size => this gives a near instant feedback regardless of the haystack size and we get an actually pretty nice cancellation mechanism out of it.
- I added a couple other memory optimizations which I found along the way (linux+text goes from ~11GB to ~7GB of RAM)
- other more minor optimizations

I'm going to merge this as-is and we can keep dogfooding on the main branch before next release.

alexpasmantier (approved):

alexpasmantier:
Thanks for bringing up such a great PR @saghen 🎉

saghen:
Sweet, thanks for merging! On the frizbee side, I need to add a per-pattern `max_typos` config and fix some bugs with unicode I discovered over the weekend. After which, I'll open a PR for frecency and a separate one for typo resistance. Bit slammed though, so it might be a little while.

PR #1126, opened 2026-08-18
feat(themes): support "none" color for the terminal default

### LINKED ISSUE #1103: Provide a way to unset color
**Problem**
I'd like to use a builtin theme, `catppuccin`, except I'd like the background to be unset (which results in transparent in my terminal emulator), but there doesn't seem to be a way to override a color with none.

```toml
[ui]
theme = "catppuccin"
[ui.theme_overrides]
background = "..." # How to disable background??
```

**Feature**
Introduce a (pseudo) color name `"none"` (or maybe the empty string) to represent the terminal default fg/bg color. The specified element should be shown in the terminal's default color, as if no color was specified.

**Examples**
```toml
[ui]
theme = "catppuccin"
[ui.theme_overrides]
background = "none"
```

ChrisJr404:
## 📺 PR Description

This adds a `"none"` pseudo-color that maps to the terminal's default color, so a theme (or a `theme_overrides` entry) can leave an element uncolored instead of forcing a concrete color. It closes #1103.

The motivation is transparency: when you start from a builtin theme but want the background to fall through to your terminal emulator (which may be transparent), there was previously no way to express that through an override — every value had to be a concrete ANSI or hex color. `background = "none"` now keeps the terminal's own background. The same works for any foreground field if you'd rather inherit the terminal default there too.

Under the hood this is a new `Color::Reset` variant that resolves to `ratatui`'s `Color::Reset`. Parsing goes through the single `Color::from_str` used by both theme files and overrides, so both paths pick it up. `"none"` is the primary spelling; `"default"` and `"reset"` are accepted as aliases. Existing themes are unaffected — no builtin uses these words.

Docs: added a line to the theme color reference in `docs/user-guide/05-themes.md`.

## Checklist

### Tests

Added four unit tests in `config::themes`: `"none"`/`"default"`/`"reset"` (case-insensitive) parse to `Color::Reset`; `Color::Reset` converts to `ratatui::style::Color::Reset`; `background = "none"` deserializes from a theme file; and an override of `background = "none"` merges correctly. `cargo test --lib themes` passes (11/11), and `cargo fmt --check` / `cargo clippy --lib` are clean.

### DIFF
--- a/docs/user-guide/05-themes.md
+++ b/docs/user-guide/05-themes.md
@@ -71,4 +71,5 @@ match_fg = "#ff0000"
 ### Colors
 
 ANSI: `"red"`, `"bright-blue"`, `"white"`  
-Hex: `"#ff0000"`, `"#1e1e2e"`
+Hex: `"#ff0000"`, `"#1e1e2e"`  
+Terminal default: `"none"` leaves the element uncolored, e.g. `background = "none"` keeps a transparent background.
--- a/television/config/themes.rs
+++ b/television/config/themes.rs
@@ -16,6 +16,11 @@ pub mod builtin;
 pub enum Color {
     Ansi(ANSIColor),
     Rgb(RGBColor),
+    /// The terminal's default color, leaving the underlying cell untouched.
+    ///
+    /// This is useful to keep a transparent background or to fall back to the
+    /// terminal's default foreground for a given element.
+    Reset,
 }
 
 impl Color {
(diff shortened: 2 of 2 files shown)


alexpasmantier:
Hi, thanks so much for the PR!

I'd like us to keep a single value and avoid aliases: `"none"` seems to make the most sense for the end user here.

Apart from that this looks great 👍🏻

ChrisJr404:
Makes sense. Dropped the `default`/`reset` aliases so `none` is the single accepted value, and updated the themes doc to match. Case-insensitive parsing is kept (`none`/`None`/`NONE`). Thanks!

alexpasmantier (approved):

